PR #2, opened 2022-11-14
[2941][ADD] download_account_move_multiple_attachments

AungKoKoLin1997:
[2941](https://www.quartile.co/web#id=2941&action=771&model=project.task&view_type=form&menu_id=505)

After downloaded zip file, filename will be displayed like the image.
Screenshot from 2022-11-30 00-40-16


yostashiro (changes requested):
@AungKoKoLin1997 Thanks for the module.  Can we take a similar approach as how auditlog generates action records so that we do not need to create a module per model?

https://github.com/oca/server-tools/blob/2f38b2eaa32c1b4ce3fd9bb82b448151627b8797/auditlog/models/rule.py#L695-L713

If this idea is valid, then I'd suggest the name to be attachment_batch_download.

AungKoKoLin1997:
> Can we take a similar approach as how auditlog generates action records so that we do not need to create a module per model?

I don't think we can take this approach. If we choose the business document model, we will not download attachments from all records. So, we need to choose records and I think it's impossible to create a one2many field that is valid for every model(I mean if we choose sale.order model, we need to select sale order records and otherwise). And what we want is to download from business documents.

AungKoKoLin1997:
This PR has already created for targeting 16.0.

PR #4, opened 2022-12-09
[2941][ADD] attachment_batch_download

AungKoKoLin1997:
[2941](https://www.quartile.co/web#id=2941&action=771&model=project.task&view_type=form&menu_id=505)

After installing this module, you need to generate download model in setting. 
316940384_1188569255374389_2719351836805843530_n

317779839_888529015633407_8875531703874198435_n

After generated download action for model, you can download attachments in document list view.

317354464_1241636766699159_4497268329652886785_n

After downloaded zip file, filename will be displayed like the image.
204611923-e1b005b9-5d34-48a4-8d50-58adf09297f4


AungKoKoLin1997:
I think we may propose this module to OCA.

PR #21, opened 2023-04-14
[3384][ADD] stock_picking_product_type

rinaldifirdaus:
[3384](https://www.quartile.co/web?db=quartile&token=dMRr78CMkEX6gyy9nTiv#id=3384&action=771&model=project.task&view_type=form&menu_id=505)

In current code, I am trying to hide the product type on picking code other than "incoming".
But the result is, the field is still shown only value is hidden.
I will discuss with client whether the product type is exclusively want to be shown only in Reception or it's ok to be displayed on the other picking code.


stock_move_product_type/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module displays product type in lines of stock move.
```
yostashiro:
```suggestion
This module displays product type in stock moves in the picking form.
```

yostashiro (commented):
Can we change the module name to stock_picking_product_type?

yostashiro:
> I will discuss with client whether the product type is exclusively want to be shown only in Reception or it's ok to be displayed on the other picking code.

I would personally go for the simpler option of showing the field in all pickings first when I am not convinced with the reason to do otherwise.

rinaldifirdaus:
@yostashiro well noted. I will let the field and its value to be shown in all picking. and will follow up if there's any feedback or additional request from client.  Thank you

yostashiro (approved):

PR #28, opened 2023-05-29
[3326][ADD] account_move_line_analytic_search

rinaldifirdaus:
[3326](https://www.quartile.co/web#id=3326&action=771&model=project.task&view_type=form&menu_id=505)


yostashiro:
@rinaldifirdaus I find it a bit tedious to prepare a module like this for all the related views.  Do you have any concerns with taking the approach of defining necessary filters using https://github.com/qrtl/axls-oca/tree/16.0/base_custom_filter?

Another option I thought was to prepare a generic module (say, analytic_distribution_search) to add the search definition in the filter in case analytic_distribution is not already included in it, but I guess base_custom_filter is reasonable enough.  Your thoughts?

rinaldifirdaus:
Closing this PR, client using base_custom_filter for this inquiry.

PR #30, opened 2023-06-22
[3422][16.0][FIX] mrp_subcontracting_lot_price_unit, add tests

AungKoKoLin1997:
[3422](https://www.quartile.co/web?debug=1#id=3422&action=771&model=project.task&view_type=form&menu_id=505)

Before this commit, the case where a product is received once, with a specific lot and a unit price of 1000, the behavior of stock_lot_price_unit should maintain this price when the same product, with the same lot, is received again, even if the new unit price is 2000. This is to respect the existing value of the unit price associated with that particular lot.

However, when the module mrp_subcontracting_lot_price_unit is installed, this behavior changes. The module will not maintain the existing value if the second receipt of the product is not a subcontracted batch. Instead, it will update the existing unit price of the lot with the new value. In this case, it would update the original unit price of 1000 with the new unit price of 2000.


mrp_subcontracting_lot_price_unit/models/stock_move_line.py
```diff
@@ -16,6 +16,6 @@ def _skip_lot_price_unit_update(self):
         subcontracting manufacting order.
         """
         res = super()._skip_lot_price_unit_update()
-        if res and self.move_id.is_subcontract:
+        if res or self.move_id.is_subcontract:
             return True
         return False
```
yostashiro:
```suggestion
        if self.move_id.is_subcontract:
            return True
        return res
```

stock_lot_price_unit/tests/test_stock_lot_price_unit.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.stock.tests.common import TestStockCommon
+
+
+class TestStockLotPriceUnit(TestStockCommon):
+    @classmethod
+    def setUpClass(cls):
+        super(TestStockLotPriceUnit, cls).setUpClass()
```
yostashiro:
```suggestion
        super().setUpClass()
```

mrp_subcontracting_lot_price_unit/tests/test_mrp_subcontracting_lot_price_unit.py
```diff
@@ -0,0 +1,35 @@
...
+
+from odoo.addons.stock_lot_price_unit.tests.test_stock_lot_price_unit import (
+    TestStockLotPriceUnit,
+)
+
+
+class MrpSubcontractingLotPriceUnit(TestStockLotPriceUnit):
+    @classmethod
+    def setUpClass(cls):
+        super(MrpSubcontractingLotPriceUnit, cls).setUpClass()
```
yostashiro:
```suggestion
        super().setUpClass()
```

yostashiro (commented):
@AungKoKoLin1997 Thanks for catching the issue.

mrp_subcontracting_lot_price_unit/tests/test_mrp_subcontracting_lot_price_unit.py
```diff
@@ -0,0 +1,35 @@
...
+        move = self.create_stock_move(picking, self.productB, 1000)
+        self.create_stock_move_line(move, self.lot1)
+
+        picking.action_confirm()
+        picking.action_assign()
+
+        move.is_subcontract = True
+        picking._action_done()
+        # Check if the lot's price_unit is not changed
+        self.assertEqual(self.lot1.price_unit, 50.0)
```
yostashiro: Not a good scenario, since `_skip_lot_price_unit_update()` of stock_lot_price_unit returns True anyway.

mrp_subcontracting_lot_price_unit/tests/test_mrp_subcontracting_lot_price_unit.py
```diff
@@ -0,0 +1,35 @@
...
+        self.lot.price_unit = 00.0
```
yostashiro: What is this line about?

stock_lot_price_unit/tests/test_stock_lot_price_unit.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def create_picking(self):
+        picking = self.env["stock.picking"].create(
+            {
+                "location_id": self.supplier_location,
+                "location_dest_id": self.stock_location,
+                "picking_type_id": self.picking_type_in,
+            }
+        )
+        return picking
```
yostashiro:
```suggestion
        return self.env["stock.picking"].create(
            {
                "location_id": self.supplier_location,
                "location_dest_id": self.stock_location,
                "picking_type_id": self.picking_type_in,
            }
        )
```

stock_lot_price_unit/tests/test_stock_lot_price_unit.py
```diff
@@ -0,0 +1,91 @@
...
+                "product_id": product.id,
+                "location_id": picking.location_id.id,
+                "location_dest_id": picking.location_dest_id.id,
+                "product_uom": product.uom_id.id,
+                "product_uom_qty": 5.0,
+                "picking_id": picking.id,
+                "price_unit": price,
+            }
+        )
+        return move
```
yostashiro:
```suggestion
        return self.env["stock.move"].create(
            {
                "name": "Move product to stock",
                "product_id": product.id,
                "location_id": picking.location_id.id,
                "location_dest_id": picking.location_dest_id.id,
                "product_uom": product.uom_id.id,
                "product_uom_qty": 5.0,
                "picking_id": picking.id,
                "price_unit": price,
            }
        )
```

yostashiro (changes requested):
I think we might as well remove the test of mrp_subcontracting_lot_price_unit if it's too involved.

yostashiro (approved):

PR #31, opened 2023-06-26
[3515][ADD] stock_move_accounting_date

rinaldifirdaus:
[3515](https://www.quartile.co/web#id=3515&action=771&model=project.task&view_type=form&menu_id=505)


stock_move_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,21 @@
...
+        store=True,
+    )
+
+    @api.depends("accounting_date")
+    def _compute_accounting_date(self):
+        for line in self:
+            if not line.picking_id.accounting_date:
+                line.accounting_date = line.picking_id.date_done
+            else:
+                line.accounting_date = line.picking_id.accounting_date
```
yostashiro:
```suggestion
        for line in self:
            line.accounting_date = line.picking_id.date_done
            if line.picking_id.accounting_date:
                line.accounting_date = line.picking_id.accounting_date
```

@rinaldifirdaus Please double check to see if we really need to look at `date_done` of the picking.  Is there no field we can use in `stock.move`?

stock_move_accounting_date/models/stock_move.py
```diff
@@ -0,0 +1,21 @@
...
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    accounting_date = fields.Date(
+        compute="_compute_accounting_date",
+        store=True,
+    )
+
+    @api.depends("accounting_date")
```
yostashiro:
```suggestion
    @api.depends("picking_id.date_done", "picking_id.accounting_date")
```

yostashiro (changes requested):

stock_move_accounting_date/models/stock_move.py
rinaldifirdaus: there's a "date" field on stock move which showing scheduled date of stock picking if the picking not done yet, and showing effective date (date_done).
as per discussed on meet room, im going to change it to the "date" field instead of date_done from picking.

yostashiro (commented):
@rinaldifirdaus Please add test-requirements.txt to clear the test error (pay attention to the CI status).

stock_move_accounting_date/models/stock_move.py
yostashiro: > there's a "date" field on stock move which showing scheduled date of stock picking if the picking not done yet, and showing effective date (date_done).

Beware that this may not the case with stock moves from MOs (please check the behavior), and we might have to handle this separately upon confirming the need with the client.

rinaldifirdaus: @yostashiro san, yes i checked the behavior and confirm that the stock move of the product (not the component) from MOs has no done date so in the stock move menu the date remain the same with the date_planned_start.

yostashiro (approved):
@rinaldifirdaus Please regenerate readme.

yostashiro (approved):

PR #35, opened 2023-07-05
[3533][ADD] stock_move_ignore_last_purchase_date

AungKoKoLin1997:
[3533](https://www.quartile.co/web#view_type=form&model=project.task&id=3533&active_id=3533&menu_id=)


AungKoKoLin1997:
@rinaldifirdaus  Can you please review first? Please suggest me about readme if my understanding is wrong.

stock_move_ignore_last_purchase_date/models/product_product.py
```diff
@@ -0,0 +1,30 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    @api.depends("stock_move_ids.state", "man_last_purchase_date")
+    def _compute_last_purchase_date(self):
```
yostashiro: @AungKoKoLin1997 Please avoid a complete override of methods.  Make the original method extendable by splitting the relevant part into another method.

product_last_purchase_date/models/product_product.py
```diff
@@ -20,16 +20,21 @@ class Product(models.Model):
...
     @api.depends("stock_move_ids.state", "man_last_purchase_date")
     def _compute_last_purchase_date(self):
         for product in self:
             last_purchase_date = False
             man_last_purchase_date = product.man_last_purchase_date
             move = product.stock_move_ids.filtered(
-                lambda m: m.state == "done" and m.picking_code == "incoming"
+                lambda m: m.state == "done"
+                and m.picking_code == "incoming"
+                and isinstance(m.id, int)
```
yostashiro: Can you please comment why this change is needed.

AungKoKoLin1997:
There is one issue. That is when we update man_last_purchase_date, the error will be raised but date will changed. The issue is from not returning actual id in one2many field. https://github.com/odoo/odoo/issues/127523

stock_move_ignore_last_purchase_date/models/product_product.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    def _assign_last_purchase_date(self, move):
+        last_purchase_date = super()._assign_last_purchase_date(move)
+        if move.ignore_last_purchase_date and self.man_last_purchase_date:
+            last_purchase_date = self.man_last_purchase_date
+        return last_purchase_date
```
yostashiro:
```suggestion
        if move.ignore_last_purchase_date:
            return False
        return super()._assign_last_purchase_date(move)
```

yostashiro (changes requested):

stock_move_ignore_last_purchase_date/models/stock_move.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    ignore_last_purchase_date = fields.Boolean(
+        related="picking_id.ignore_last_purchase_date", store=True
```
yostashiro: Updating this field doesn't automatically trigger the update of `last_purchase_date` of the product.  We may make this a compute field, and call `_compute_last_purchase_date()` inside the compute method.

rinaldifirdaus (approved):
Functional test: works as expected

stock_move_ignore_last_purchase_date/models/product_product.py
```diff
@@ -0,0 +1,25 @@
...
+    def _assign_last_purchase_date(self, move):
+        if move.ignore_last_purchase_date:
+            new_move = self.stock_move_ids.filtered(
+                lambda m: m.state == "done"
+                and m.picking_code == "incoming"
+                and not m.ignore_last_purchase_date
+            ).sorted(key=lambda m: m.id, reverse=True)[:1]
+            if new_move:
+                move_date = fields.Date.context_today(self, new_move.date)
+                if move_date > self.man_last_purchase_date:
```
kanda999:
```suggestion
                if self.man_last_purchase_date and move_date > self.man_last_purchase_date:
```
If nothing is entered for `man_last_purchase_date`, it is a comparison of datetime and false so I get a error.

kanda999 (changes requested):

kanda999:
@AungKoKoLin1997 
I have updated this PR Could you review this please?. It probably has the requirements of the module.
Your method of using `assign_last_purchase_date` seemed be not good.  This is because the `_compute_last_purchase_date` method already uses filtered to identify the move. However, passing the identified move as an argument to a new method and then filtering the move again would result in two unnecessary filters.

What we should do is filter moves that do not include the `ignore_last_purchase_date` that was added this stock_move_ignore_purchase_last_date module, so we should create the domain as a new method (`_get_domain`) and extend that domain.

If you have a concern, let me know...!

AungKoKoLin1997:
@kanda999  Thanks for the update. Having `_get_domain` function is good and more clear code. I agree with your approach.

product_last_purchase_date/models/product_product.py
```diff
@@ -38,3 +38,10 @@ def _compute_last_purchase_date(self):
                 last_purchase_date = man_last_purchase_date
             product.last_purchase_date = last_purchase_date
             product.product_tmpl_id._compute_last_purchase_date()
+
+    def _get_domain(self, product):
```
yostashiro:
```suggestion
    def _get_move_domain(self, product):
```
Just to reduce the chances of the method name clash...

We usually have place the method above the caller, for better readability.

stock_move_ignore_last_purchase_date/views/stock_picking_views.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="name">stock.picking.form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <field name="origin" position="after">
+                <field
+                    name="ignore_last_purchase_date"
+                    attrs="{'invisible': [('picking_type_code', '!=', 'incoming')]}"
+                />
+            </field>
```
yostashiro: @rinaldifirdaus The field doesn't have to be (shouldn't really be, IMO) in such a noticeable place, although users may not case much about it.  Please check with the client on this point.

rinaldifirdaus: @AungKoKoLin1997 , client agree to move the position of this field into the other info tab. could you please help to update?thanks

yostashiro (approved):

stock_move_ignore_last_purchase_date/views/stock_move_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_move_tree" model="ir.ui.view">
+        <field name="name">stock.move.tree</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_move_tree" />
+        <field name="arch" type="xml">
+            <xpath expr="//tree/field[@name='state']" position="after">
+                <field name="ignore_last_purchase_date" />
```
yostashiro:
```suggestion
                <field name="ignore_last_purchase_date" optional="hide" />
```
Consider to assign `optional` attribute whenever we add fields to list views.

rinaldifirdaus (approved):

kanda999 (approved):
LGTM 👍

PR #43, opened 2023-08-24
[16.0][3504][ADD] purchase_next_reception_date

AungKoKoLin1997:
[3504](https://www.quartile.co/web#id=3504&cids=3&model=project.task&view_type=form)


AungKoKoLin1997:
@rinaldifirdaus  Please review this behavior and also add background in readme.

rinaldifirdaus:
>Please review this behavior and also add background in readme.

Hi @AungKoKoLin1997 , i've added some changes and add the background.
could you please help to review it? thanks!

purchase_next_reception_date/models/purchase_order.py
```diff
@@ -12,12 +12,18 @@ class PurchaseOrder(models.Model):
...
+        "order_line.date_planned",
+        "order_line.product_qty",
+        "order_line.qty_received",
+        "order_line.qty_invoiced",
     )
     def _compute_next_reception_date(self):
         for order in self:
             dates = order.order_line.filtered(
-                lambda line: line.product_qty != line.qty_received
+                lambda line: line.product_id.detailed_type != "service"
```
AungKoKoLin1997, line 23: I don't understand why we need to pay attention about service product. I believe next_reception_date should be only for receivable products. For the service product PO, next_reception_date should not be set, isn't?

purchase_next_reception_date/models/purchase_order.py
```diff
@@ -0,0 +1,29 @@
...
+        "order_line.qty_invoiced",
+    )
+    def _compute_next_reception_date(self):
+        for order in self:
+            dates = order.order_line.filtered(
+                lambda line: line.product_id.detailed_type != "service"
+                and line.qty_received < line.product_qty
+                or line.product_id.detailed_type == "service"
+                and line.qty_invoiced < line.product_qty
+            ).mapped("date_planned")
```
AungKoKoLin1997: I think this is better way. Please check it.
```
dates = order.order_line.filtered(
                lambda line: (
                    (line.product_id.detailed_type != "service" and line.qty_received < line.product_qty) 
                    or 
                    (line.product_id.detailed_type == "service" and line.qty_invoiced < line.product_qty)
                )
            ).mapped("date_planned")
```

purchase_next_reception_date/views/purchase_order_views.xml
yostashiro, line 1: @rinaldifirdaus We should add the field in the search views (for RFQs and orders) as well, both filter and group by.

yostashiro (changes requested):

rinaldifirdaus:
@yostashiro @AungKoKoLin1997 
Thank you for the review!
i've just added the new commit on this PR.
i add the next_reception_date field on the RFQ list view and set it optional hide. (I assume our client  mostly need information of this field on the purchase order list view, adding this just in case they need in RFQ they can set this field to be displayed)

purchase_next_reception_date/views/purchase_order_views.xml
```diff
@@ -0,0 +1,75 @@
...
+                <filter
+                    string="Next Reception Date"
+                    name="next_reception_date"
+                    context="{'group_by':'next_reception_date'}"
+                />
+            </group>
+        </field>
+    </record>
+    <record id="purchase_order_view_search_inherit" model="ir.ui.view">
+        <field name="name">purchase.order.inherit.purchase.order.search</field>
```
yostashiro: Let's uncomplicate our practice - just use the same id and name as the inhrited record.

```suggestion
    <record id="purchase_order_view_search" model="ir.ui.view">
        <field name="name">purchase.order.select</field>
```

purchase_next_reception_date/views/purchase_order_views.xml
```diff
@@ -0,0 +1,75 @@
...
+                />
+            </filter>
+            <group expand="0" string="Group By">
+                <filter
+                    string="Next Reception Date"
+                    name="next_reception_date"
+                    context="{'group_by':'next_reception_date'}"
+                />
+            </group>
+        </field>
```
yostashiro:
```suggestion
            <xpath expr="//filter[@name='order_date']" position="after">
                <filter
                    name="next_reception_date"
                    string="Next Reception Date"
                    date="next_reception_date"
                />
            </xpath>
            <xpath expr="//group" position="inside">
                <filter
                    string="Next Reception Date"
                    name="next_reception_date"
                    context="{'group_by':'next_reception_date'}"
                />
            </xpath>
        </field>
```

Same for `view_purchase_order_filter`.

purchase_next_reception_date/models/purchase_order.py
```diff
@@ -0,0 +1,35 @@
...
+
+    next_reception_date = fields.Date(
+        compute="_compute_next_reception_date", store=True
+    )
+
+    @api.depends(
+        "order_line.date_planned",
+        "order_line.product_qty",
+        "order_line.qty_received",
+        "order_line.qty_invoiced",
```
yostashiro: Add `product_id` just in case.

```suggestion
        "order_line.product_id",
        "order_line.product_qty",
        "order_line.qty_received",
        "order_line.qty_invoiced",
        "order_line.date_planned",
```

purchase_next_reception_date/models/purchase_order.py
```diff
@@ -0,0 +1,35 @@
...
+                        and line.qty_received < line.product_qty
+                    )
+                    or (
+                        line.product_id.detailed_type == "service"
+                        and line.qty_invoiced < line.product_qty
+                    )
+                )
+            ).mapped("date_planned")
+
+            order.next_reception_date = min(dates) if dates else False
```
yostashiro: Non-blocking minor preference.

```suggestion
            lines = order.order_line.filtered(
                lambda x: (
                    (
                        x.product_id.detailed_type != "service"
                        and x.qty_received < x.product_qty
                    )
                    or (
                        x.product_id.detailed_type == "service"
                        and x.qty_invoiced < x.product_qty
                    )
                )
            )
            dates = lines.mapped("date_planned")
            order.next_reception_date = min(dates) if dates else False
```

purchase_next_reception_date/readme/DESCRIPTION.rst
yostashiro, line 1: Please review and fix the grammar.

purchase_next_reception_date/views/purchase_order_views.xml
```diff
@@ -0,0 +1,65 @@
...
+    </record>
+    <record id="view_purchase_order_filter" model="ir.ui.view">
+        <field name="name">request.quotation.select</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.view_purchase_order_filter" />
+        <field name="arch" type="xml">
+            <xpath expr="//filter[@name='order_date']" position="after">
+                <filter name="next_reception_date" date="next_reception_date" />
+            </xpath>
+            <xpath expr="//group" position="inside">
```
yostashiro: A non-blocking minor suggestion just to be a bit more consistent with the filter (i.e. next_receiption_date shows right after order_date).

```suggestion
            <xpath expr="//group/filter[@name='order_date']" position="after">
```

purchase_next_reception_date/views/purchase_order_views.xml
```diff
@@ -0,0 +1,65 @@
...
+    </record>
+    <record id="purchase_order_view_search" model="ir.ui.view">
+        <field name="name">purchase.order.select</field>
+        <field name="model">purchase.order</field>
+        <field name="inherit_id" ref="purchase.purchase_order_view_search" />
+        <field name="arch" type="xml">
+            <xpath expr="//filter[@name='order_date']" position="after">
+                <filter name="next_reception_date" date="next_reception_date" />
+            </xpath>
+            <xpath expr="//group" position="inside">
```
yostashiro: Same as above.

purchase_next_reception_date/models/purchase_order.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class PurchaseOrder(models.Model):
+    _inherit = "purchase.order"
+
+    next_reception_date = fields.Date(
```
yostashiro: I just realized that this field is defined as `Date` and it will cause an issue, for example, when the next receipt date according to the order lines is expected to be '2023-08-27 08:00:00' JST - the next reception date in the header will be '2023-08-26'.

Therefore, we should either make the field `Datetime`, or convert the datetime into date with `context_timestamp()` in the compute method.  From design perspective, we should go for the former, as the result of the latter will depend on user's timezone setting.

yostashiro (changes requested):

rinaldifirdaus (approved):

yostashiro (approved):
Code review.

PR #44, opened 2023-09-11
[3757][16.0][IMP] inventory_reporting: add unbuild sheet

AungKoKoLin1997:
[3757](https://www.quartile.co/web#id=3757&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)

### DIFF
--- a/inventory_reporting/reports/inventory_report.py
+++ b/inventory_reporting/reports/inventory_report.py
@@ -87,48 +87,37 @@ class InventoryReportXlsx(models.AbstractModel):
                 row += 1
 
     def generate_storable_report(self, workbook, wizard):
+        base_storable_domain = [
+            ["accounting_date", ">=", wizard.date_start],
+            ["accounting_date", "<=", wizard.date_end],
+            ["product_id.active", "=", True],
+            ["product_id.detailed_type", "=", "product"],
+        ]
         categories = [
             {
                 "name": _("Receipt"),
                 "filter": [
-                    ["accounting_date", ">=", wizard.date_start],
-                    ["accounting_date", "<=", wizard.date_end],
                     ["stock_move_id.picking_type_id.code", "=", "incoming"],
                     ["stock_move_id.origin_returned_move_id", "=", False],
-                    ["product_id.active", "=", True],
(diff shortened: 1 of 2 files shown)


rinaldifirdaus (approved):
Thanks lin. the report of unbuild operation generated properly.

inventory_reporting/reports/inventory_report.py
```diff
@@ -173,6 +174,16 @@ def generate_storable_report(self, workbook, wizard):
...
                 ],
             },
+            {
+                "name": _("Unbuild"),
+                "filter": [
+                    ["accounting_date", ">=", wizard.date_start],
+                    ["accounting_date", "<=", wizard.date_end],
+                    ["product_id.active", "=", True],
+                    ["product_id.detailed_type", "=", "product"],
+                    ["stock_move_id.unbuild_id", "!=", False],
```
yostashiro: @rinaldifirdaus @AungKoKoLin1997 Why don't you make a common method to return the base domain (or define the base domain at the beginning of the method) to which you make adjustments depending on the report type?  We shouldn't be repeating the same conditions several times.

inventory_reporting/reports/inventory_report.py
```diff
@@ -87,48 +87,37 @@ def generate_valuation_report(self, workbook, wizard):
                 row += 1
 
     def generate_storable_report(self, workbook, wizard):
+        base_storable_domain = [
+            ["accounting_date", ">=", wizard.date_start],
+            ["accounting_date", "<=", wizard.date_end],
+            ["product_id.active", "=", True],
+            ["product_id.detailed_type", "=", "product"],
+        ]
```
yostashiro: @AungKoKoLin1997 Is there a reason why this is done with lists (not with tuples)? If no specific reason, we should use tuples to go with the convention, and also to avoid unnecessary hiccups (not sure some methods like `expression.AND` works with domains with lists).

yostashiro (approved):
Code review.

kanda999 (approved):
Functional review:  Unbuild sheets could be printed.

PR #55, opened 2023-10-19
[3438][3894][IMP] account_move_obc_csv

AungKoKoLin1997:
[3438](https://www.quartile.co/web#id=3438&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)
[3894](https://www.quartile.co/web#id=3894&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)

* Add demarkator column ("GL0010000") in the output file to indicate the first line of an
  OBC journal entry (stock journal entries from pickings and MOs are grouped into one
  OBC journal entry picking/MO).
* Show the partner code of the subcontracted partner in the output file (in "GL0012009")
  for journal envries from subcontracting MOs.

### DIFF
--- a/account_move_obc_csv/report/report_account_move_obc_csv.py
+++ b/account_move_obc_csv/report/report_account_move_obc_csv.py
@@ -14,7 +14,7 @@ class AccountMoveObcCsv(models.AbstractModel):
 
     def _get_fields(self):
         return [
-            "GL0010000",  # asterisk
+            "GL0010000",  # demarkator: indicates the start of a new OBC record
             "GL0010001",  # date
             "GL0012002",  # debit account
             "GL0012101",  # debit base amount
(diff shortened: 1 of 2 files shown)


account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -177,12 +179,24 @@ def _get_report_vals_dict(self, record):
...
+        ):
+            first_record = True
+            for record in grouped_records:
+                vals_dict = self._get_report_vals_dict(record)
+                if first_record:
+                    vals_dict[1]["GL0010000"] = "*"
+                    first_record = False
+                for _k, v in sorted(vals_dict.items()):
+                    writer.writerow(v)
+                record.is_exported = True
```
yostashiro: We need to consider the non-inventory journal entries (e.g. those from vendor bills).  The first line need to be marked with `*` for those as well.
```suggestion
        picking = self.env["stock.picking"]
        for record in sorted_records:
            vals_dict = self._get_report_vals_dict(record)
            first_record = False
            record_picking = record.stock_move_id.picking_id
            if not record_picking:
                first_record = True
            elif picking != record_picking:
                first_record = True
                picking = record_picking
            if first_record:
                vals_dict[1]["GL0010000"] = "*"
            for _k, v in sorted(vals_dict.items()):
                writer.writerow(v)
            record.is_exported = True
```

yostashiro (approved):
Should be good, but we haven't added consideration to consolidate journal entries of manufacturing orders.  @rinaldifirdaus Please check with the client if this should be covered in this update.

rinaldifirdaus:
>Should be good, but we haven't added consideration to consolidate journal entries of manufacturing orders.  Please check with the client if this should be covered in this update.

Hi @AungKoKoLin1997 , client request to add the consideration to consolidate the JE of Manufacturing order as well. could you please help to update this PR to handle this request? thanks!

AungKoKoLin1997:
>  client request to add the consideration to consolidate the JE of Manufacturing order as well

If I am not mistaken, there is no partner_id field in MO. So, when we create MO, the generated stock_move and JE will not be assigned partner_id. Customer wants which value to be partner_ref for MO?

rinaldifirdaus:
@AungKoKoLin1997 , sorry to confuse you, but this request is about giving "*" only for the first line of the same transfer/mo. currently all line on the report of same MO is having "*" . could you please check the behavior?

yostashiro:
@AungKoKoLin1997 I have updated the code to slightly improve the readability by flipping the conditions, and also to fix an issue with record sorting - production ids need to be considered or the journal entries from the same MO can be split apart.

@rinaldifirdaus Please go ahead and apply the change to the staging server.

PR #56, opened 2023-10-20
[16.0][3863][ADD] crm_exception

AungKoKoLin1997:
I will consider creating PR in OCA after the design is approved.
[3863](https://www.quartile.co/web#id=3863&cids=3&model=project.task&view_type=form)


rinaldifirdaus (approved):
IMO, we can go ahead with this design. as workaround, the client should prepare multiple exception for each field and give the warning message on exception name such "Please fill the Phone field", so that the users will acknowledge of which field that is required to be filled.

crm_exception/models/crm_lead.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def crm_check_exception(self):
+        self._check_exception()
+
+    def _check_crm_lead_check_exception(self, vals):
+        check_exceptions = any(
+            field in vals for field in self._fields_trigger_check_exception()
+        )
+        if check_exceptions:
+            self.crm_check_exception()
```
yostashiro:
```suggestion
    def _check_crm_lead_check_exception(self, vals):
        check_exceptions = any(
            field in vals for field in self._fields_trigger_check_exception()
        )
        if check_exceptions:
            self._check_exception()
```

crm_exception/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Crm Exception",
```
yostashiro:
```suggestion
    "name": "CRM Exception",
```

crm_exception/__manifest__.py
```diff
@@ -0,0 +1,13 @@
...
+    "version": "16.0.1.0.0",
+    "category": "Tools",
```
yostashiro:
```suggestion
    "category": "Customer Relationship Management",
```

crm_exception/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows you attach several customizable exceptions to your opportunity.
```
yostashiro: Check grammar.

crm_exception/views/crm_lead_views.xml
```diff
@@ -0,0 +1,17 @@
...
+<odoo>
+    <record id="crm_lead_view_form" model="ir.ui.view">
+        <field name="name">crm.lead.form.inherit</field>
+        <field name="model">crm.lead</field>
+        <field name="inherit_id" ref="crm.crm_lead_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='tag_ids']" position="after">
+                <field
+                    name="ignore_exception"
+                    states="purchase"
```
yostashiro: Propose an alternative or simply remove it.

yostashiro (changes requested):

crm_exception/models/crm_lead.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def write(self, vals):
```
yostashiro, line 26: @AungKoKoLin1997 Any reason to use `write()` instead of `@api.constrains()`?  Shouldn't the latter make the code more concise?

AungKoKoLin1997: The main purpose is to avoid hardcoding a trigger field list in @api.constrains and ensure flexibility for potential future additions to the _fields_trigger_check_exception. May be rare case but I think it is also good for design.

yostashiro: @AungKoKoLin1997 Isn't that practically the same as the trigger fields are hardcoded anyway in `_fields_trigger_check_exception()` or am I missing something?

AungKoKoLin1997: Yes. The trigger fields are hardcoded in `_fields_trigger_check_exception()` and we may extend this function and can add new more fields from another module if we will need some adjustments in the future.

yostashiro (commented):
@AungKoKoLin1997 Can you provide the reason why we are now not supporting the `ignore_exception` option?

AungKoKoLin1997:
> Can you provide the reason why we are now not supporting the `ignore_exception` option?

I misunderstood. I will revert it.

AungKoKoLin1997:
Closing this PR. I will create in OCA repo.

PR #57, opened 2023-10-27
[3921][FIX] product_plm_import: remove other encoding options than utf-8

yostashiro:
[3921](https://www.quartile.co/web#id=3921&cids=3&model=project.task&view_type=form)

There was a corner-case issue of data dcoding by cp932 wouldn't throw an exception when utf-8 file has been imported, resulting in rumbled characters in the imported content.

This commit fixes the issue by removing the other options (cp932 and utf-8-bom) than utf-8 to eliminate such kind of possibilities, as the imported file should always be encoded in utf-8.


rinaldifirdaus (approved):
Functional test: works as expected!

PR #64, opened 2023-12-13
[4052][IMP] inventory_reporting: Add Scrap sheet and add shipping records to Delivery sheet

rinaldifirdaus:
[4052](https://www.quartile.co/web#id=4052&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)


inventory_reporting/reports/inventory_report.py
```diff
@@ -118,7 +119,13 @@ def generate_storable_report(self, workbook, wizard):
             {
                 "name": _("Delivery"),
                 "filter": [
+                    "|",
                     ("stock_move_id.picking_type_id.code", "=", "internal"),
+                    (
+                        "stock_move_id.reference",
+                        "ilike",
+                        "/SHIP/",
```
yostashiro: Not comfortable with this condition. Shouldn't we judge based on `stock_move_id.location_dest_id.usage`?

inventory_reporting/reports/inventory_report.py
```diff
@@ -135,6 +142,26 @@ def generate_storable_report(self, workbook, wizard):
                         "ilike",
                         "Virtual Locations/Inventory adjustment",
```
yostashiro: This condition also looks bad.

inventory_reporting/reports/inventory_report.py
```diff
@@ -135,6 +142,26 @@ def generate_storable_report(self, workbook, wizard):
...
+                    ),
+                ],
+            },
+            {
+                "name": _("Scrap"),
+                "filter": [
+                    (
+                        "stock_move_id.scrapped",
+                        "!=",
+                        False,
```
yostashiro:
```suggestion
                        "stock_move_id.scrapped", "=", True,
```

inventory_reporting/reports/inventory_report.py
```diff
@@ -135,6 +142,26 @@ def generate_storable_report(self, workbook, wizard):
...
+                    ),
+                    (
+                        "stock_move_id.picking_code",
+                        "!=",
+                        "internal",
+                    ),
```
yostashiro: Do we need this part?

yostashiro (commented):
@rinaldifirdaus @AungKoKoLin1997 Can you please review the filter definitions as they currently appear to be rather random/unitidy. Avoid relying on the reference values, for example, unless it's really necessary.

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: @yostashiro san, I just reconsider it and seems this is not needed in the client case. I was thinking that if the user proceed the scrap directly from the receipt transfer, the receipt transfer data will shown both in receipt sheet and scrap sheet. but since our client use the Scrap operation type to record the scrap, we can delete this lines.

By the way, since there's several request to add new sheets, as you mentioned in the past @yostashiro san, I think the design should be improved to be not hardcode the filter each sheets. maybe it can be changed to have 1 menu to provide setting form to add and define only the sheets of storable and consumable reports. for inventory valuation seems its ok with current design. how do you think @AungKoKoLin1997

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: @yostashiro san, For the Shipping (/SHIP/) operation type, the destination location that they set  is virtual location/production. i think if set stock_move_id.location_dest_id.usage = production, the subcontracting transfer will also shown in the delivery sheet, no?

rinaldifirdaus:
@yostashiro san, @AungKoKoLin1997 I have updated the conditions, kindly your review. thanks!
With previous filter, i found in the storable reports there are a records in delivery sheet which have internal transfer for scrap, inventory adjustment, and return.

with the new filter, in delivery sheet only show the transfer that sending the goods from the stock to production.
the other internal transfer for scrap, inventory adjustment, and return allocated to each sheets.

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
                     ("stock_move_id.origin_returned_move_id", "=", False),
+                    ("stock_move_id.unbuild_id", "=", False),
                 ],
             },
             {
                 "name": _("Return"),
                 "filter": [
+                    "|",
+                    ("stock_move_id.picking_type_id.code", "=", "internal"),
+                    ("stock_move_id.picking_type_id.code", "=", "outgoing"),
```
yostashiro:
```suggestion
                    ("stock_move_id.picking_code", "in", ("internal", "outgoing")),
```
Do we actually need `internal`?

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
                 "filter": [
                     ("stock_move_id.picking_type_id.code", "=", "incoming"),
                     ("stock_move_id.origin_returned_move_id", "=", False),
+                    ("stock_move_id.unbuild_id", "=", False),
                 ],
             },
             {
                 "name": _("Return"),
```
yostashiro:
```suggestion
                "name": _("Return"),  # Supplier returns
```

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
                 "filter": [
+                    "|",
+                    ("stock_move_id.picking_type_id.code", "=", "internal"),
+                    ("stock_move_id.picking_type_id.code", "=", "outgoing"),
                     ("stock_move_id.origin_returned_move_id", "!=", False),
-                    ("reference", "ilike", "/OUT/"),
                 ],
             },
             {
                 "name": _("Delivery"),
```
yostashiro:
```suggestion
                "name": _("Component Flush"),
```

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
                 "filter": [
+                    "|",
                     ("stock_move_id.picking_type_id.code", "=", "internal"),
+                    ("stock_move_id.picking_type_id.code", "=", "outgoing"),
```
yostashiro: Do we need this condition?

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
+                    ("stock_move_id.unbuild_id", "=", False),
+                    ("stock_move_id.origin_returned_move_id", "=", False),
```
yostashiro: Do we need this condition?

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
-                    ),
-                    (
-                        "stock_move_id.location_dest_id",
-                        "ilike",
-                        "Virtual Locations/Inventory adjustment",
-                    ),
+                    ("stock_move_id.location_id.usage", "!=", "production"),
+                    ("stock_move_id.location_dest_id.usage", "!=", "production"),
+                    ("stock_move_id.picking_id", "=", False),
+                    ("stock_move_id.unbuild_id", "=", False),
```
yostashiro: Why can't it just be `"stock_move_id.location_dest_id.usage", "=", "inventory"`?

inventory_reporting/reports/inventory_report.py
```diff
@@ -106,59 +106,52 @@ def generate_storable_report(self, workbook, wizard):
...
-                        "ilike",
-                        "Physical Locations/Subcontracting Location",
-                    ),
-                    (
-                        "stock_move_id.location_dest_id",
-                        "ilike",
-                        "Physical Locations/Subcontracting Location",
-                    ),
+                    ("stock_move_id.location_id.usage", "=", "production"),
+                    ("stock_move_id.location_dest_id.usage", "=", "production"),
```
yostashiro: To filter records "subcontracting".  Please check the validity of this suggestion.

```suggestion
                    "&",
                    ("stock_move_id.location_id.usage", "=", "production"),
                    ("stock_move_id.location_dest_id.is_subcontracting_location", "=", True),
                    "&",
                    ("stock_move_id.location_dest_id.usage", "=", "production"),
                    ("stock_move_id.location_id.is_subcontracting_location", "=", True),
```

yostashiro (commented):
@rinaldifirdaus Please compare the outputs before and after the change to make sure that there is no regression.

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: Yes, I think for return mainly they are do return directly from the transfer of the related reception. but if i check on their DB, sometimes they also using Internal transfer operation to do the return.

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: This is to get the shipping records (/SHIP/)

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: for the unbuild_id is false, this is to address the records that generated when return the subcontracting reception with the module mrp_unbuild_subcontracting behavior.

inventory_reporting/reports/inventory_report.py
rinaldifirdaus: Ok Tashiro san, i will check with this.

rinaldifirdaus:
@yostashiro san, have added the new commit. i will add this improvement, and let the customer check. and will let them know that at the moment this  code is under review. thank you

rinaldifirdaus:
@yostashiro san, I did comparison the data of the previous filter and this new filter (data from June to Dec of copy-DB). and the data completely same. thank you

rinaldifirdaus:
@yostashiro san, can we have this PR merged? thank you

rinaldifirdaus:
Hi @liuhehe1995 , i've just add one commit to update a "supplier return" based on yesterday discussion with Tashiro san and this version is the one im confirming with the client on the email. you can compare and see the result between report generated by this version, and the one with report generated on the production db. thank you

rinaldifirdaus:
@yostashiro san & @liuhehe1995 san, kindly review the latest commit. i compared the data from production with data from dev env using this update, and the result was as expected (subcontracting return should on subcontract sheet, and inventory adjustment even the operation type is internal transfer should be recorded on inventory adjustment). 
I will apply this change to staging and will let customer check the behavior. thank you

yostashiro (approved):

PR #68, opened 2024-01-30
[4233][ADD] mrp_unbuild_owner

AungKoKoLin1997:
[4233](https://www.quartile.co/web#id=4233&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)


mrp_unbuild_restore_original_quant/models/mrp_unbuild.py
```diff
@@ -0,0 +1,55 @@
...
+                        self.env['stock.move.line'].create({
+                            'move_id': move.id,
+                            'qty_done': move.product_uom_qty,
+                            'product_id': move.product_id.id,
+                            'product_uom_id': move.product_uom.id,
+                            'location_id': move.location_id.id,
+                            'location_dest_id': move.location_dest_id.id,
+                        })
+                        move.write({'state': 'confirmed'})
+                        moves += move
```
yostashiro: Not getting the point of this `else` part. Shouldn't we just use the standard logic in this case?

mrp_unbuild_restore_original_quant/models/mrp_unbuild.py
```diff
@@ -0,0 +1,63 @@
...
+                                        "product_uom_id": move.product_uom.id,
+                                        "location_id": move.location_id.id,
+                                        "location_dest_id": move.location_dest_id.id,
+                                    }
+                                )
+                            move.write({"state": "confirmed"})
+                        moves += move
+                else:
+                    moves += super(MrpUnbuild, unbuild)._generate_produce_moves()
+            moves = moves.with_context(produce_move=True)
```
yostashiro: @AungKoKoLin1997 I forget why we are assigning this context. Can you please explain?

yostashiro:
@AungKoKoLin1997 I've added a commit.  Please check if this makes sense.

mrp_unbuild_restore_original_quant/models/mrp_unbuild.py
```diff
@@ -0,0 +1,59 @@
...
+    def _prepare_move_line_vals(self, move, origin_move_line, taken_quantity):
+        vals = super()._prepare_move_line_vals(move, origin_move_line, taken_quantity)
+        if origin_move_line.owner_id:
+            vals["owner_id"] = origin_move_line.owner_id.id
+        return vals
+
+    def action_unbuild(self):
+        self.ensure_one()
+        if self.mo_id:
+            self = self.with_context(exact_unbuild=True)
```
AungKoKoLin1997: @yostashiro  I think this will skip `action_confirm()` for consumed_moves.
https://github.com/odoo/odoo/blob/d8ec7735aa8759fa94fc0223945ac9b19e05312e/addons/mrp/models/mrp_unbuild.py#L167-L168

mrp_unbuild_restore_original_quant/models/mrp_unbuild.py
AungKoKoLin1997: I added this because I want to skip `action_confirm`() for only `produce_moves`. For `consume_moves`, I want to work as standard. May be we need to handle it later but now it is not inside of current scope.

yostashiro (approved):

yostashiro:
@AungKoKoLin1997 Can you please create a PR in oca/manufacture in your spare time? Priority is not high at this point as the design is a bit clumsy until https://github.com/odoo/odoo/pull/155751 gets processed.

PR #71, opened 2024-02-14
[16.0][3489][ADD] auth_oauth_force_oauth

AungKoKoLin1997:
[3489](https://www.quartile.co/web#id=3489&cids=3&model=project.task&view_type=form)


yostashiro (commented):
Module name suggestion: auth_oauth_block_password_login

Haven't checked the code.

auth_oauth_block_password_login/controllers/main.py
```diff
@@ -0,0 +1,99 @@
...
+class OAuthLogin(Login):
+    @http.route()
+    def web_login(self, *args, **kw):
+        ensure_db()
+        response = super(OAuthLogin, self).web_login(*args, **kw)
+        if response.is_qweb:
+            error = request.params.get("oauth_error")
+            if error == "4":
+                error = _(
+                    "You are not allowed to login with Oauth. Please login with password."
```
kanda999: There is no need to restrict other users to password login only.

It is sufficient if users with domains are restricted to only oauth.

auth_oauth_block_password_login/controllers/main.py
```diff
@@ -0,0 +1,99 @@
...
+                )
+                response.qcontext["error"] = error
+        return response
+
+
+class OAuthController(OAuthController):
```
kanda999: I was thinking that domain users would be just redirected to web/login when they access web/reset_password.

But if you have concerns about that, this logic is fine.

auth_oauth_block_password_login/controllers/main.py
```diff
@@ -0,0 +1,99 @@
...
+        return super().signin(**kw)
+
+
+class Home(home.Home):
+    @http.route("/web/login", type="http", auth="none")
+    def web_login(self, redirect=None, **kw):
+        # Before proceeding with the standard login process, check conditions
+        if request.httprequest.method == "POST":
+            login = kw.get("login")
+            if login:
```
yostashiro: Can we get rid of these if ladders?

auth_oauth_block_password_login/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    allow_oauth_domains = fields.Text()
```
yostashiro: `force_oauth_domains` is probably a more suitable name.  Please explain the domains should be separated in the field (by comma, by line break or something else).

auth_oauth_block_password_login/controllers/main.py
```diff
@@ -0,0 +1,99 @@
...
+                    return redirect
+        return super().signin(**kw)
+
+
+class Home(home.Home):
+    @http.route("/web/login", type="http", auth="none")
+    def web_login(self, redirect=None, **kw):
+        # Before proceeding with the standard login process, check conditions
+        if request.httprequest.method == "POST":
+            login = kw.get("login")
```
yostashiro: Is there no simpler way than doing all the following steps?

auth_oauth_block_password_login/controllers/main.py
yostashiro: We don't need this if I'm not mistaken. There should be nothing blocking the use of OAuth login.

yostashiro (changes requested):
Another name suggestion after doing a quick review of the code: `auth_oauth_force_oauth`

Not perfectly sure if the logic we implement will be blocking other means of login than password (SAML, LDAP, etc.), so the module name may depend on the module design.  @AungKoKoLin1997 @kanda999 What do you think?

kanda999 (approved):
Functional review: It worked as expected.

auth_oauth_force_oauth/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Auth Oauth Force Oauth",
+    "version": "16.0.1.0.0",
+    "author": "Quartile Limited",
+    "website": "https://www.quartile.co",
+    "category": "Oauth",
```
yostashiro:
```suggestion
    "category": "Tools",
```

auth_oauth_force_oauth/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Auth Oauth Force Oauth",
```
yostashiro:
```suggestion
    "name": "Auth OAuth Force OAuth",
```

auth_oauth_force_oauth/controllers/main.py
```diff
@@ -0,0 +1,59 @@
...
+        password = qcontext.get("password")
+        confirm_password = qcontext.get("confirm_password")
+        if not login or not password or not confirm_password:
+            return super(CustomAuthSignup, self).web_auth_reset_password(*args, **kw)
+        user = request.env["res.users"].sudo().search([("login", "=", login)], limit=1)
+        force_domains = user.company_id.force_oauth_domains
+        if not force_domains:
+            return super(CustomAuthSignup, self).web_auth_reset_password(*args, **kw)
+        if not any(login.endswith(domain) for domain in force_domains):
+            return super(CustomAuthSignup, self).web_auth_reset_password(*args, **kw)
```
yostashiro: The style looks a bit different to the corresponding part of `Home.web_login()`.  Make the style consistent, and I suggest that we create a method `_is_allowed_password_login()` in `res.users` and call the method to avoid repeating the same logic.

```suggestion
        if user._is_allowed_password_login()
            return super().web_auth_reset_password(*args, **kw)
```

yostashiro (changes requested):

auth_oauth_force_oauth/models/res_users.py
```diff
@@ -0,0 +1,19 @@
...
+    @api.model
+    def _is_allowed_password_login(self):
+        force_domains = self.company_id.force_oauth_domains
+        if not force_domains:
+            return True
+        # Check if the login ends with any of the force domains
+        force_domains_list = [domain.strip() for domain in force_domains.split(",")]
+        if not any(self.login.endswith(domain) for domain in force_domains_list):
+            return True
+        return False
```
yostashiro:
```suggestion
        return not any(self.login.endswith(domain) for domain in force_domains_list):
```

auth_oauth_force_oauth/controllers/main.py
```diff
@@ -0,0 +1,51 @@
...
+        if not login or not password or not confirm_password:
+            return super().web_auth_reset_password(*args, **kw)
+        user = request.env["res.users"].sudo().search([("login", "=", login)], limit=1)
+        if user._is_allowed_password_login():
+            return super().web_auth_reset_password(*args, **kw)
+        qcontext["error"] = _(
+            "You are not allowed to login with password. Please use OAuth login."
+        )
+        response = request.render("auth_signup.reset_password", qcontext)
+        return response
```
yostashiro:
```suggestion
        return request.render("auth_signup.reset_password", qcontext)
```

auth_oauth_force_oauth/models/res_company.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    force_oauth_domains = fields.Text(help="Fill in the domains, separated by commas.")
```
yostashiro: Forgot to ask earlier. Why should this field be `Text` if domains should be separated by commas?

yostashiro (commented):
@AungKoKoLin1997 FYI, no need to squash commits unless we do want to keep specific commits in a PR (e.g. migration) in the qrtl repos, as we do it when we do the merge.

auth_oauth_force_oauth/models/res_company.py
AungKoKoLin1997: You right. It should be `Char` field.

yostashiro (approved):
Code review.

@kanda999 Please double-check the function in the test database before pushing this to production.

kanda999:
Double-checked functionality.

Please be careful not to remove the commits for the translations I have added. @AungKoKoLin1997

PR #74, opened 2024-03-13
[4063][ADD] stock_picking_location_check

AungKoKoLin1997:
[4063](https://www.quartile.co/web#id=4063&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)


stock_picking_location_check/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    allow_location_discrepancy = fields.Boolean(
+        help="If enable, the location discrepancies will be allowed between picking "
+        "and stock move lines."
+    )
+
+    @api.constrains("state")
```
yostashiro: Shouldn't we use `location_id`, `location_dest_id`, `move_line_ids.location_id`, and `move_line_ids.location_dest_id`, instead?

stock_picking_location_check/models/stock_picking.py
```diff
@@ -0,0 +1,37 @@
...
+        """Check consistency of locations between picking and its stock move lines."""
+        for picking in self:
+            if picking.allow_location_discrepancy:
+                continue
+            # Determine if the picking is an immediate transfer
+            is_immediate_transfer = picking._check_immediate()
+            for move_line in picking.move_line_ids:
+                # If not an immediate transfer and qty_done is 0, skip this move line
+                if not is_immediate_transfer and move_line.qty_done <= 0:
+                    continue
```
yostashiro: In what situation would we get into trouble without this?

stock_picking_location_check/views/stock_picking_views.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_picking_form" model="ir.ui.view">
+        <field name="name">stock.picking.form</field>
+        <field name="model">stock.picking</field>
+        <field name="inherit_id" ref="stock.view_picking_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='picking_type_id']" position="after">
```
yostashiro:
```suggestion
            <xpath expr="//field[@name='location_id']" position="before">
```

stock_picking_location_check/views/stock_picking_views.xml
```diff
@@ -0,0 +1,16 @@
...
+                <field
+                    name="allow_location_discrepancy"
```
yostashiro:
```suggestion
                    name="allow_location_discrepancy"
                    groups="stock.group_stock_multi_locations"
```

stock_picking_location_check/views/stock_picking_views.xml
```diff
@@ -0,0 +1,16 @@
...
+                    attrs="{'readonly': [('state', '=', 'done')]}"
```
yostashiro:
```suggestion
                    attrs="{'readonly': [('state', 'in', ('done', 'cancel'))]}"
```

stock_picking_location_check/models/stock_picking.py
AungKoKoLin1997: If we test with below situation, the validation error will be raised.
1. Create planned transfer for receipt for lot product 
2. Add move lines with the destination location that is different from picking's destination location.
3. Then, you change picking's dest location that is the same with move lines location.

Then, you will see like two move lines with different dest locations. As per odoo standard behavior, when we changes location in picking, it also changes to related moves locations and move lines location but doesn't change to `0.00` qty lines.
Screenshot 2024-03-14 at 23 46 35

stock_picking_location_check/models/stock_picking.py
AungKoKoLin1997: Unless I misunderstood about the behavior of `@api.constrains` decorator, it is only triggered when these fields are included in the create and write call. So, when we change location_dest_id or location_id of move lines from detailed operations, these function will not trigger.

kanda999 (approved):
Functional review: LGTM

yostashiro:
@AungKoKoLin1997 I've added a commit. Please check the validity of it. I've switched to extend _action_done instead of adding a constraint method, to avoid checks at unexpected/unnecessary timings.

AungKoKoLin1997:
Superseded by https://github.com/qrtl/axls-oca/pull/134.

PR #78, opened 2024-04-09
[4414][FIX] stock_move_line_move_location

kanda999:
[4414](https://www.quartile.co/web#id=4414&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)

Fix there is only one transfer line proposed when I seleccted two stock move lines. 
2024-04-09_12h06_58
(one stock move line with owner and one stock move line without owner)

### DIFF
--- a/stock_move_line_move_location/wizard/stock_move_location.py
+++ b/stock_move_line_move_location/wizard/stock_move_location.py
@@ -37,8 +37,8 @@ class StockMoveLocationWizard(models.TransientModel):
             return super()._prepare_wizard_move_lines(move_lines)
         # if need move only available qty per product on location
         for _product, ml in groupby(
-            sorted(move_lines, key=lambda r: (r.product_id, r.lot_id)),
-            lambda r: (r.product_id, r.lot_id),
+            sorted(move_lines, key=lambda r: (r.product_id, r.lot_id, r.owner_id)),
+            lambda r: (r.product_id, r.lot_id, r.owner_id),
         ):
             quant = self.env["stock.quant"]
             ml = list(ml)[0]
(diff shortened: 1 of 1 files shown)


kanda999:
@AungKoKoLin1997 
I guess you were referring to this part of the stock_move_line module.
https://github.com/qrtl/axls-oca/blob/2a1e6749abb68d187033c17bcc11fa7c4d6923b7/stock_move_location/wizard/stock_move_location.py#L116

Do we need to group stock_move_line as well? 
I think no need for that.

AungKoKoLin1997:
> Do we need to group stock_move_line as well?
> I think no need for that.

@kanda999 Yes. It is needed and I did for intentionally. Why should we show multiple lines in wizard for the lines that have same product_id and lot_id?

AungKoKoLin1997:
As per this fix, you should not remove groups part and should add owner_id in grouping parameters.

stock_move_line_move_location/wizard/stock_move_location.py
```diff
@@ -37,15 +37,17 @@ def _prepare_wizard_move_lines(self, move_lines):
             return super()._prepare_wizard_move_lines(move_lines)
         # if need move only available qty per product on location
         for _product, ml in groupby(
-            sorted(move_lines, key=lambda r: (r.product_id, r.lot_id)),
-            lambda r: (r.product_id, r.lot_id),
+            sorted(move_lines, key=lambda r: (r.product_id, r.lot_id, r.owner_id)),
+            lambda r: (r.product_id, r.lot_id, r.owner_id),
```
yostashiro: Any reason `r.package_id` is not included?

yostashiro (approved):
Code review.

kanda999:
Functional review: LGTM

PR #83, opened 2024-06-21
[4121][ADD] stock_picking_validate_jobcan_wf

AungKoKoLin1997:
Superseding of https://github.com/qrtl/axls-private/pull/49
[4121](https://www.quartile.co/web#id=4121&cids=3&model=project.task&view_type=form)


AungKoKoLin1997:
I can't decide we should display the `ValidationError` or keep current design that is displayed warning message in related picking.

Downside of current approach:
- When we validate batch records and some pickings are not validated because of jobcan workflow is not completed, the `ValidationError` is not raised and user may think all are successfully validated.

Downside of displaying `ValidationError`:
- When user validates the pickings and ValidationError will be raised. So, the picking that should be successful can't be validated. Then, when user opens the record of picking that is failed because of jobcan workflow, user can't know he didn't validate or failed for some reason. So, he have to click validate button again.

yostashiro:
> I can't decide we should display the `ValidationError` or keep current design that is displayed warning message in related picking.
> 
> Downside of current approach:
> 
> * When we validate batch records and some pickings are not validated because of jobcan workflow is not completed, the `ValidationError` is not raised and user may think all are successfully validated.
> 
> Downside of displaying `ValidationError`:
> 
> * When user validates the pickings and ValidationError will be raised. So, the picking that should be successful can't be validated. Then, when user opens the record of picking that is failed because of jobcan workflow, user can't know he didn't validate or failed for some reason. So, he have to click validate button again.

Took the approach of showing UserError in my update. There is almost no benefit in partially processing pickings in batch validation for the added complexity, when the batch validation is triggered by an operator (not by some automated process).

yostashiro (approved):

PR #87, opened 2024-07-17
[4720][IMP] stock_picking_validate_jobcan_wf: skip check jobcan ID for retu…

AungKoKoLin1997:
[4720](https://www.quartile.co/web#id=4720&cids=3&model=project.task&view_type=form)

### DIFF
--- a/stock_picking_validate_jobcan_wf/models/stock_picking.py
+++ b/stock_picking_validate_jobcan_wf/models/stock_picking.py
@@ -13,14 +13,10 @@ class StockPicking(models.Model):
     _name = "stock.picking"
     _inherit = ["stock.picking", "api.call.mixin"]
 
-    skip_jobcan_wf = fields.Boolean("Skip Jobcan Workflow")
-    show_skip_jobcan_wf = fields.Boolean(
-        compute="_compute_show_skip_jobcan_wf", store=True
-    )
+    skip_jobcan_wf = fields.Boolean("Skip Jobcan Workflow", copy=False)
+    show_skip_jobcan_wf = fields.Boolean(compute="_compute_show_skip_jobcan_wf")
     jobcan_wf_number = fields.Char("Jobcan Workflow Number", copy=False)
-    show_jobcan_wf_number = fields.Boolean(
-        compute="_compute_show_jobcan_wf_number", store=True
-    )
+    show_jobcan_wf_number = fields.Boolean(compute="_compute_show_jobcan_wf_number")
 
     def _is_outgoing(self):
         self.ensure_one()
@@ -37,25 +33,21 @@ class StockPicking(models.Model):
             return True
         return False
 
-    @api.depends("picking_type_id", "location_id", "location_dest_id", "move_ids")
     def _compute_show_skip_jobcan_wf(self):
         for pick in self:
+            pick.show_skip_jobcan_wf = False
             if (
-                not pick._is_outgoing()
-                or pick._receipt_return_picking()
-                or pick.move_line_ids.mapped("owner_id")
+                pick._is_outgoing()
+                and not pick._receipt_return_picking()
+                and not pick.move_line_ids.mapped("owner_id")
             ):
-                pick.show_skip_jobcan_wf = False
-                pick.skip_jobcan_wf = True
-            else:
                 pick.show_skip_jobcan_wf = True
-                pick.skip_jobcan_wf = False
 
     @api.depends("skip_jobcan_wf")
     def _compute_show_jobcan_wf_number(self):
         for pick in self:
             pick.show_jobcan_wf_number = False
-            if not pick.skip_jobcan_wf:
+            if pick.show_skip_jobcan_wf and not pick.skip_jobcan_wf:
                 pick.show_jobcan_wf_number = True
 
     def get_api_key(self, config):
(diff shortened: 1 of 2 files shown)


stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -36,7 +49,11 @@ def _compute_show_skip_jobcan_wf(self):
     def _compute_show_jobcan_wf_number(self):
         for pick in self:
             pick.show_jobcan_wf_number = False
-            if pick._is_outgoing() and not pick.skip_jobcan_wf:
+            if (
+                pick._is_outgoing()
+                and not pick.skip_jobcan_wf
+                and not pick._receipt_return_picking()
```
yostashiro: Why do we need this adjustment?
```suggestion
```

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -25,9 +25,22 @@ def _is_outgoing(self):
...
         ] and self.location_dest_id.usage not in ["internal", "transit"]
 
+    def _receipt_return_picking(self):
+        self.ensure_one()
+        if not self.move_ids:
+            return False
+        move = self.move_ids[0]
+        if move.origin_returned_move_id and move._is_out():
+            return True
+        return False
```
yostashiro: I think the code can be leaner. Technically speaking, Jobcan WF should be skipped if ALL items in the outgoing transfer are returns to the vendor.

yostashiro (changes requested):

stock_picking_validate_jobcan_wf/models/stock_picking.py
AungKoKoLin1997: This adjustment is needed for not showing and requiring of `jobcan_wf_number` for receipts return.
https://github.com/qrtl/axls-custom/blob/3720a67608535eea4fdf2b787340386c8f26f041/stock_picking_validate_jobcan_wf/views/stock_picking_views.xml#L15-L18

stock_picking_validate_jobcan_wf/models/stock_picking.py
AungKoKoLin1997: This adjustment is needed for not showing and requiring of `jobcan_wf_number` for receipts return.
https://github.com/qrtl/axls-custom/blob/3720a67608535eea4fdf2b787340386c8f26f041/stock_picking_validate_jobcan_wf/views/stock_picking_views.xml#L15-L18

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -25,9 +25,22 @@ def _is_outgoing(self):
...
+
     def _compute_show_skip_jobcan_wf(self):
         for pick in self:
-            if not pick._is_outgoing() or pick.move_line_ids.mapped("owner_id"):
+            if (
+                not pick._is_outgoing()
+                or pick.move_line_ids.mapped("owner_id")
+                or pick._receipt_return_picking()
+            ):
                 pick.show_skip_jobcan_wf = False
```
yostashiro:
```suggestion
            if (
                not pick._is_outgoing()
                or pick._receipt_return_picking()
                or pick.move_line_ids.mapped("owner_id")
            ):
                pick.show_skip_jobcan_wf = False
                pick.skip_jobcan_wf = True
```

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -36,7 +49,11 @@ def _compute_show_skip_jobcan_wf(self):
     def _compute_show_jobcan_wf_number(self):
         for pick in self:
             pick.show_jobcan_wf_number = False
-            if pick._is_outgoing() and not pick.skip_jobcan_wf:
+            if (
+                pick._is_outgoing()
+                and not pick.skip_jobcan_wf
+                and not pick._receipt_return_picking()
+            ):
```
yostashiro:
```suggestion
            if not pick.skip_jobcan_wf:
```

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -25,9 +25,22 @@ def _is_outgoing(self):
...
         ] and self.location_dest_id.usage not in ["internal", "transit"]
 
+    def _receipt_return_picking(self):
+        self.ensure_one()
+        if not self.move_ids:
+            return False
+        for move in self.move_ids:
+            if not (move.origin_returned_move_id and move._is_out()):
+                return False
+        return True
```
yostashiro: Non-blocking. I guess it doesn't make much difference.
```suggestion
        if self.move_ids and self.move_ids == self.move_ids.filtered(
            lambda x: x._is_out() and x.origin_returned_move_id
        ):
            return True
        return False
```

yostashiro (commented):
@AungKoKoLin1997 How about my new suggestion?

stock_picking_validate_jobcan_wf/models/stock_picking.py
AungKoKoLin1997: Adding `pick.skip_jobcan_wf = True` and make `skip_jobcan_wf` as a compute field is not worked as expected behavior because this compute function is shared for both store and non store field. Can we just keep previous design because sharing compute function for store and non-store fields is not also a good practice from technical POV?

yostashiro:
@AungKoKoLin1997 Can you please check my last commit.

yostashiro (approved):

AungKoKoLin1997:
> Can you please check my last commit.

@yostashiro  I checked the functionals and it works well. Thank you.

yostashiro:
@kanda999 Please go ahead and take care of the rest.

PR #88, opened 2024-08-02
[4750][IMP] product_last_purchase_date

AungKoKoLin1997:
[4750](https://www.quartile.co/web#id=4750&cids=3&model=project.task&view_type=form)

### DIFF
--- a/product_last_purchase_date/models/product_product.py
+++ b/product_last_purchase_date/models/product_product.py
@@ -22,7 +22,7 @@ class Product(models.Model):
 
     @api.depends(
         "stock_move_ids.state",
-        "stock_move_ids.account_move_ids.date",
+        "stock_move_ids.account_move_ids.state",
         "man_last_purchase_date",
     )
     def _compute_last_purchase_date(self):
@@ -33,9 +33,13 @@ class Product(models.Model):
                 self._get_move_domain(product), order="id desc", limit=1
             )
             if move:
-                last_purchase_date = fields.Date.context_today(self, move.date)
-                if move.account_move_ids:
-                    last_purchase_date = move.account_move_ids[0].date
+                posted_moves = move.account_move_ids.filtered(
+                    lambda x: x.state == "posted"
+                )
+                if posted_moves:
+                    last_purchase_date = posted_moves[0].date
+                else:
+                    last_purchase_date = fields.Date.context_today(self, move.date)
             if (
                 not last_purchase_date
                 or man_last_purchase_date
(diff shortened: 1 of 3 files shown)


product_last_purchase_date/models/product_product.py
```diff
@@ -29,7 +29,7 @@ def _compute_last_purchase_date(self):
                 self._get_move_domain(product), order="id desc", limit=1
             )
             if move:
-                last_purchase_date = fields.Date.context_today(self, move.date)
+                last_purchase_date = move.accounting_date
```
kanda999: If the accounting_date is not set, won't anything be updated?

kanda999:
Is it possible to make the logic to get the accounting_date of the account.move related with the stock.move?
I noticed that this way we don't have to add more dependency.

product_last_purchase_date/models/product_product.py
AungKoKoLin1997: With the design of computing accounting_date, it will always have value.
https://github.com/qrtl/axls-oca/blob/4085c38bc89d12ee9afa1194283d16191b4d00c8/stock_picking_accounting_date/models/stock_move.py#L18-L21

AungKoKoLin1997:
> Is it possible to make the logic to get the accounting_date of the account.move related with the stock.move?

This is not valid logic because every move doesn't have related account_move if the product is not real_time valuation costing method.

kanda999:
@AungKoKoLin1997 
I think it's fine to use the logic of checking the accounting_date for real_time products. 
We only want to base it on the accounting_date for real_time products

AungKoKoLin1997:
> We only want to base it on the accounting_date for real_time products

So, we don't want to set last_purchase_date for not real_time products?
My intention is not to add conditions for using accounting_date of account.move or date of stock.move for the clean code.
I think adding dependency is not huge maintenance cost if we prefer the code quality.

kanda999:
@AungKoKoLin1997 
I thought it would be better to remove the dependency for the following two reasons:

- The dependency can be removed.
- Using the accounting_date of account.move provides more accurate information. 
This is because users might change the accounting_date of account.move later, and in such cases, we want to update the last_purchase_date of the product.

Additionally, the condition is not complicated, so adding it will not affect maintainability. 
Therefore, I think the logic should be that real-time products are based on the accounting_date, while other products use the date of stock.move.

yostashiro (approved):
Code review.

kanda999 (approved):

PR #90, opened 2024-08-07
[4721][IMP] stock_picking_validate_jobcan_wf: add schedule action to confir…

AungKoKoLin1997:
[4721](https://www.quartile.co/web#id=4721&cids=3&menu_id=517&model=project.task&view_type=form)

### DIFF
--- a/stock_picking_validate_jobcan_wf/models/stock_picking.py
+++ b/stock_picking_validate_jobcan_wf/models/stock_picking.py
@@ -1,4 +1,5 @@
 # Copyright 2024 Quartile
+# Copyright 2024 Axelspace Corporation (https://axelspace.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
 
 import logging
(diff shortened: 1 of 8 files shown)


WangTKurata (approved):

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -56,40 +57,100 @@ def get_api_key(self, config):
...
+        )
+        if not channel:
+            channel = self.env["mail.channel"].create(
+                {
+                    "name": channel_name,
+                    "channel_type": "chat",
+                    "channel_partner_ids": [(4, self.user_id.partner_id.id)],
+                }
+            )
+        return channel
```
yostashiro: @WangTKurata This will end up cluttering the discuss screen as errors accumulate.  I suppose we don't need to create a channel per picking as long as the person in charge of the picking is notified?

We might as well do something similar to this: https://github.com/oca/ocb/blob/541b400/addons/mail_bot/models/res_users.py#L25-L38

yostashiro (commented):
I have refactored the code a bit in my last commit.

@WangTKurata Any particular reason errors should appear as chat messages?  Just curious because we didn't take this approach before (i.e. product_plm_import).

stock_picking_validate_jobcan_wf/models/stock_picking.py
WangTKurata: Yes. Notified to the person in charge is OK.

yostashiro: @AungKoKoLin1997 Can you please follow up on this to adjust the cost, removing the channel creation logic?

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -56,40 +57,90 @@ def get_api_key(self, config):
...
+                "message": message,
+            }
+            subject = _("Picking Confirmation Failed: %(name)s") % {"name": self.name}
+            channel.message_post(
+                body=message_body,
+                subject=subject,
+                message_type="comment",
+                subtype_xmlid="mail.mt_comment",
+                partner_ids=[self.user_id.partner_id.id],
+            )
```
yostashiro: @AungKoKoLin1997 Can you please help switch to add a message in the document instead of in a channel.  Kurata-san agrees that it's better to make it visible in the document.  No need to manipulate with document followers.

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -96,11 +96,6 @@ def _check_jobcan_wf_status(self):
     def notify_user(self, message):
         self.ensure_one()
         if self.user_id:
```
yostashiro: We should remove this condition as we are no longer relying on this field.

```suggestion
```

yostashiro (changes requested):

stock_picking_validate_jobcan_wf/models/stock_picking.py
yostashiro: We should also adjust the method name. Change it to `jobcan_validate_error_message_post()` or similar perhaps?

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -56,40 +57,85 @@ def get_api_key(self, config):
...
+                "wf_id": self.jobcan_wf_number,
+                "url": self_url,
+                "name": self.name,
+                "message": message,
+            }
+            subject = _("Picking Confirmation Failed: %(name)s") % {"name": self.name}
+            self.message_post(
+                body=message_body,
+                subject=subject,
+                message_type="comment",
```
yostashiro: Isn't `notification`, which is the default, more appropriate?

stock_picking_validate_jobcan_wf/models/stock_picking.py
```diff
@@ -56,40 +57,85 @@ def get_api_key(self, config):
...
+                subtype_xmlid="mail.mt_note",
+                partner_ids=[self.user_id.partner_id.id],
```
yostashiro: I believe this can be removed.  Just deliver the message to the document followers.
```suggestion
```

yostashiro (approved):

PR #93, opened 2024-09-05
[4121][FIX] stock_picking_validate_jobcan: show_jobcan_wf_number

yostashiro:
[4121](https://www.quartile.co/web#id=4121&cids=3&model=project.task&view_type=form)

Before this commit, visibility of jobcan_wf_number field would depend on allocation of stock (the field would get hidden when owner stock was allocated), which was incorrect.


yostashiro:
Fast tracking.

PR #100, opened 2024-11-01
[4936][FIX] inventory_reporting: depends on stock_move_actual_date

AungKoKoLin1997:
[4936](https://www.quartile.co/web#model=project.task&id=4936&active_id=4936&cids=3)


yostashiro (approved):
Reviewing this made me think that we should incorporate the functionality of stock_valuation_layer_accounting_date into stock_move_actual_date, deprecating the former module. This can be done in the downstream versions.

yostashiro:
By the way, I see no point of basing the consumable report on stock.valuation.layer. Should have based it on stock.move (or stock.move.line, depending on how other reports might be restructured...) from the beginning.

kanda999 (approved):

PR #104, opened 2024-11-19
[4976][ADD] stock_move_line_price_unit

AungKoKoLin1997:
[4976](https://www.quartile.co/web#id=4976&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)


stock_move_line_price_unit/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Quartile
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
```
yostashiro:
```suggestion
# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl)
```

yostashiro (approved):
Code review.

stock_move_line_price_unit/i18n/ja.po
```diff
@@ -0,0 +1,26 @@
...
+
+#. module: stock_move_line_price_unit
+#: model:ir.model,name:stock_move_line_price_unit.model_stock_move_line
+msgid "Product Moves (Stock Move Line)"
+msgstr "製品の移動(在庫移動ライン)"
+
+#. module: stock_move_line_price_unit
+#: model:ir.model.fields,field_description:stock_move_line_price_unit.field_stock_move_line__price_unit
+msgid "Unit Price"
+msgstr "購買単価"
```
yostashiro: We might just make it "単価" as the value is also set for component consumptions in manufacturing, in which case "購買" is not technically correct. We may instead add a help text to the field to explain how/when the value is updated (we could use the tooltip function for this, if that makes the process easier).

```suggestion
msgstr "単価"
```

kanda999 (approved):
LGTM

PR #107, opened 2025-02-12
[4947][ADD] purchase_stock_price_variance

AungKoKoLin1997:
[4947](https://www.quartile.co/web#id=4947&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)


inventory_price_check/models/product_template.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    price_discrepancy_threshold_type = fields.Selection(
+        [("percentage", "Percentage"), ("fixed", "Fixed Value"), ("ignore", "Ignore")]
+    )
+    price_discrepancy_threshold_value = fields.Float(
+        help="Threshold value for price discrepancy warnings."
+    )
```
yostashiro: Can we make it like this:
```suggestion
    price_variance_threshold_percent = fields.Float(help="Maximum variance (in percent) allowable between the product's standard price and purchase receipt unit price.")
    price_variance_threshold_amount = fields.Float(help="Maximum allowable variance (in monetary amount, based on company currency) between the product's standard price and the purchase receipt unit price.")
```

The idea is to allow both types of thresholds. Hitting either one should trigger an error.

inventory_price_check/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, fields, models
+from odoo.exceptions import UserError
+
+
+class Stockpick(models.Model):
+    _inherit = "stock.picking"
+
+    allow_price_inconsistency = fields.Boolean(
```
yostashiro:
```suggestion
    bypass_price_variance_check = fields.Boolean(
```

inventory_price_check/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+        groups="inventory_price_check.group_purchase_inventory_discrepancy",
+        help="If enabled, no error is raised for price inconsistency between "
+        "recieved price and standard price",
+    )
+
+    def _action_done(self):
+        for pick in self:
+            if pick.picking_type_id.code != "incoming":
+                continue
+            if pick.sudo().allow_price_inconsistency:
```
yostashiro: Why `sudo()` here?

inventory_price_check/models/stock_picking.py
```diff
@@ -0,0 +1,65 @@
...
+                continue
+            for move in pick.move_ids_without_package:
+                product = move.product_id
+                threshold_type = (
+                    product.price_discrepancy_threshold_type
+                    or pick.company_id.price_discrepancy_threshold_type
```
yostashiro: Values from the company should be kept in variables outside the loop.

yostashiro (changes requested):
Partial review.

`inventory_price_check` doesn't seem to follow the normal naming convention. How about `purchase_stock_price_variance` as the module name?

inventory_price_check/models/stock_picking.py
AungKoKoLin1997: This field is only allowed for some group and we will get access right error if the user is not in this group.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+from odoo.exceptions import UserError
+
+
+class Stockpick(models.Model):
+    _inherit = "stock.picking"
+
+    bypass_price_variance_check = fields.Boolean(
+        copy=False,
+        tracking=True,
+        groups="purchase_stock_price_variance.group_bypass_price_variance_check",
```
yostashiro: Wouldn't it be better to extend write() to check if the user belongs to this group? I think this group assignment to the field is too rigid.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+                    continue
+                threshold_percent = (
+                    product.price_variance_threshold_percent
+                    or global_price_variance_threshold_percent
+                )
+                threshold_amount = (
+                    product.price_variance_threshold_amount
+                    or global_price_variance_threshold_amount
+                )
+                received_price = move.price_unit
```
yostashiro: I think it should be safer this way to update the cost based on the current exchange rate.
```suggestion
                received_price = move._get_price_unit()
```

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        global_price_variance_threshold_amount = (
+            self.env.company.price_variance_threshold_amount
+        )
+        for pick in self:
+            if pick.picking_type_id.code != "incoming":
+                continue
+            if pick.sudo().bypass_price_variance_check:
+                continue
+            for move in pick.move_ids_without_package:
```
yostashiro: Any reason for not using `pick.move_ids`?

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,60 @@
...
+    def _action_done(self):
+        global_price_variance_threshold_percent = (
+            self.env.company.price_variance_threshold_percent
+        )
+        global_price_variance_threshold_amount = (
+            self.env.company.price_variance_threshold_amount
+        )
+        for pick in self:
+            if pick.picking_type_id.code != "incoming":
+                continue
```
yostashiro: Can you please check if adding conditions using `_is_in()` and `_is_dropshipped()` for each stock move record is more appropriate instead of this?

purchase_stock_price_variance/views/product_product_views.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record id="product_normal_form_view_inherit" model="ir.ui.view">
+        <field name="name">product.product.form.inherit</field>
+        <field name="model">product.product</field>
+        <field name="inherit_id" ref="product.product_normal_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='traceability']" position="after">
+                <group string="Price Discrepancy Settings" name="price_discrepancy">
+                    <field name="price_variance_threshold_percent" />
+                    <field name="price_variance_threshold_amount" />
+                    <field name="bypass_price_variance_check" />
```
yostashiro:
```suggestion
                    <field name="bypass_price_variance_check" />
                    <field name="price_variance_threshold_percent" />
                    <field name="price_variance_threshold_amount" />
```
and make the threshold fields invisible when `bypass_price_variance_check` is true.

yostashiro (commented):
Partial review.

purchase_stock_price_variance/models/stock_picking.py
AungKoKoLin1997: When I checked the odoo source code, checking `picking_type_id.code` is used in `stock`.`_is_in()` and `_is_dropshipped()` is introduced in `stock_account` and used them. In our module, `stock_account `is part of dependency because we use `purchase_stock `as a dependency. So, I think using `_is_in()` and `_is_dropshipped()` is more appropriate.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+                percentage_difference = (
+                    (amount_difference / standard_price) * 100 if standard_price else 0
+                )
+                if (
+                    percentage_difference > threshold_percent
+                    or amount_difference > threshold_amount
+                ):
+                    raise UserError(
+                        _(
+                            f"Price discrepancy detected for {product.name}: "
```
yostashiro:
```suggestion
                            f"Price variance exceeding a threshold detected for {product.name}: "
```

AungKoKoLin1997:
I did `bypass_price_variance_check` to be `True` as a default for two reasons. First is to avoid the test cases failed of other module because the default threshold values are 0.0(it means there should be no variance) and reduce the work for the user to disable the check for all product and to set up the threshold value for the products that is needed to use for threshold check.

purchase_stock_price_variance/readme/CONFIGURE.md
yostashiro, line 1: Please tidy the steps (with numbers, perhaps) and "correct" the line breaks.

purchase_stock_price_variance/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,2 @@
+This module checks for variance between the receipt price and the product's
+standard price at the time of validation.
```
yostashiro:
```suggestion
This module checks the variance between the purchase price and the product's
standard price at the time of receipt picking validation and displays an error if the
variance exceeds a given threshold.
```

purchase_stock_price_variance/readme/USAGE.md
```diff
@@ -0,0 +1,6 @@
+A warning message will be displayed during validation if the variance
+between the receipt price and the standard price exceeds the threshold.
+
+Selecting 'Bypass Price Variance Check' allows users to bypass this error,
+and it is only visible to members of the 'Bypass Price Variance Check'
+group.
```
yostashiro:
```suggestion
An error message will be displayed during validation if the variance
between the purchase price and the standard price exceeds a threshold.

Selecting 'Bypass Price Variance Check' in the receipt picking allows users
to bypass this error. Only members of the 'Bypass Price Variance Check'
group can update this field.
```

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _action_done(self):
+        global_price_variance_threshold_percent = (
+            self.env.company.price_variance_threshold_percent
+        )
+        global_price_variance_threshold_amount = (
+            self.env.company.price_variance_threshold_amount
+        )
+        for pick in self:
+            if pick.sudo().bypass_price_variance_check:
```
yostashiro:
```suggestion
            if pick.bypass_price_variance_check:
```

yostashiro:
> I did `bypass_price_variance_check` to be `True` as a default for two reasons. First is to avoid the test cases failed of other module because the default threshold values are 0.0(it means there should be no variance) and reduce the work for the user to disable the check for all product and to set up the threshold value for the products that is needed to use for threshold check.

@AungKoKoLin1997 Sorry, missed this comment. I think the default of the product should be false, and have another field in res.config.settings to enable the feature.

yostashiro (approved):
Code review.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,76 @@
...
+                received_price = move._get_price_unit()
+                standard_price = product.standard_price
+                amount_difference = abs(received_price - standard_price)
+                percentage_difference = (
+                    (amount_difference / standard_price) * 100 if standard_price else 0
+                )
+                if (
+                    percentage_difference > threshold_percent
+                    or amount_difference > threshold_amount
+                ):
```
nobuQuartile: @AungKoKoLin1997 
I'm doing the functional test.
Currently, two problems arise.
1. The initial threshold value is 0, so if the variance check is set to true and other thresholds are not changed, all products will not be able to be received.
2. Both percentage and fixed values ​​are used for variance, making it difficult to continue adjusting those values.

When I consulted Tashiro-san, he said that when it was 0, it would be best to avoid using that threshold.
I thought so too.

I think it's a good idea to separate the process by threshold to see if they're using that value, and if they're using them, check if they're above the threshold.
What do you think?

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,75 @@
...
+                )
+                received_price = move._get_price_unit()
+                standard_price = product.standard_price
+                amount_difference = abs(received_price - standard_price)
+                percentage_difference = (
+                    (amount_difference / standard_price) * 100 if standard_price else 0
+                )
+                if (
+                    threshold_percent != 0 and percentage_difference > threshold_percent
+                ) or (threshold_amount != 0 and amount_difference > threshold_amount):
```
yostashiro:
```suggestion
                    threshold_percent and percentage_difference > threshold_percent
                ) or (threshold_amount and amount_difference > threshold_amount):
```

purchase_stock_price_variance/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+        string="Enable price variance threshold check"
+    )
+    price_variance_threshold_percent = fields.Float(
+        help="Maximum variance (in percent) allowable between the product's standard price"
+        " and purchase receipt unit price."
+    )
+    price_variance_threshold_amount = fields.Float(
+        help="Maximum allowable variance (in monetary amount, based on company currency)"
+        " between the product's standard price and the purchase receipt unit price."
+    )
```
yostashiro, line 25: Let's make sure that the values are positive here.
```suggestion
    )

    @api.constrains("price_variance_threshold_percent", "price_variance_threshold_amount")
    def _check_price_variance_threshold(self):
        for rec in self:
            if rec.price_variance_threshold_percent < 0 or rec.price_variance_threshold_amount < 0:
                raise ValidationError(_("The threshold values cannot be negative."))
```

purchase_stock_price_variance/views/res_config_setting_views.xml
```diff
@@ -0,0 +1,45 @@
...
+                    <div class="o_setting_right_pane">
+                        <label for="price_variance_threshold" />
+                        <div class="text-muted">
+                            Enable the feature to check the price variance between the purchase price unit and the product's standard price.
+                        </div>
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label for="price_variance_threshold_percent" />
+                        <div class="text-muted">
+                            Maximum variance (in percent) allowable between the product's standard price and purchase receipt unit price.
```
yostashiro:
```suggestion
                            Maximum variance (in percent) allowable between the product's standard price and purchase receipt unit price.
                            Keeping it at zero means that this threshold will not be checked.
```

Please apply the same adjustments to all the relevant places.

purchase_stock_price_variance/views/res_config_setting_views.xml
```diff
@@ -0,0 +1,45 @@
...
+                        </div>
+                        <div class="content-group">
+                            <div class="mt16">
+                                <field name="price_variance_threshold_percent" />
```
yostashiro: Not sure this is the correct way to put it. Please check it yourself, and do the same for all the relevant fields.
```suggestion
                                <field name="price_variance_threshold_percent" /><span> %</span>
```

yostashiro (changes requested):
@nobuQuartile Please remove the pre-commit noise from other modules.

nobuQuartile:
@AungKoKoLin1997 @yostashiro 
Global values ​​cannot be ignored for each threshold at the product level and cannot be allowed.

Current Product Variant Options
The initial value is set to 0
- Set a value at the product level (+)
- Disable the variant at the product level for a specific threshold (Not possible)
- Use the global value (0)

Option1
When using numeric types, the only way to divide other than 0 and plus is negative. 
- Disable the variant at the product level for a specific threshold (-)

Option2
- Set a value at the product level (+)
- Disable the variant at the product level for a specific threshold (0)
- Use the global value (-)

Option3
Create a new boolean for each threshold at the product level.
- Set a value at the product level (+)
- Disable the variant at the product level for a specific threshold (0)
- Use the global value (New boolean)

I think Option3 is better.
What do you think?

nobuQuartile:
We decided not to change the code.

Current Product Variant Options
The initial value is set to 0

Set a value at the product level (+)
Disable the variant at the product level for a specific threshold (Large value that is not typical)
Use the global value (0)

purchase_stock_price_variance/models/res_company.py
nobuQuartile: I have added this section

purchase_stock_price_variance/readme/CONFIGURE.md
nobuQuartile: I have fixed the readme.

purchase_stock_price_variance/views/res_config_setting_views.xml
nobuQuartile: I have added the comments.

purchase_stock_price_variance/views/res_config_setting_views.xml
nobuQuartile: I have added the span.

yostashiro: @nobuQuartile How about adjustments on the product views?

purchase_stock_price_variance/readme/CONFIGURE.md
```diff
@@ -0,0 +1,15 @@
+1. Navigate to Inventory > Configuration > Settings.
+2. Find and enable the "Enable Price Variance Threshold Check" option.
+3. Set the following global values to apply if no specific value is set at the product level.
+    If the threshold value is set to 0, this threshold will not be checked.:
+    * Price Variance Threshold Percent – Default percentage variance for all products.
+    * Price Variance Threshold Amount – Default maximum variance for all products.
```
yostashiro:
```suggestion
    If the threshold value is set to 0, this threshold will not be checked.
    * **Price Variance Threshold Percent**: Default percentage variance for all products.
    * **Price Variance Threshold Amount**: Default maximum variance for all products.
```

purchase_stock_price_variance/readme/CONFIGURE.md
```diff
@@ -0,0 +1,15 @@
...
+4. Go to Inventory > Products and open the product.
+5. Click on the Inventory tab and configure the following fields.
+    If the threshold value is set to 0, the threshold will refer to the global value.
+    It is not possible to disregard a single global threshold value from the product level.:
+    * Bypass Price Variance Check – Enable this to skip the check for this specific product.
+    * Price Variance Threshold Percent – Set the allowable percentage variance.
+    * Price Variance Threshold Amount – Define the maximum allowable price variance.
```
yostashiro: We should either add more explanation on the "limitation" with an example, or remove the sentence to avoid confusing the audience.
```suggestion
    * **Bypass Price Variance Check**: Enable this to skip the check for this specific product.
    * **Price Variance Threshold Percent**: Set the allowable percentage variance.
    * **Price Variance Threshold Amount**: Define the maximum allowable price variance.
```

purchase_stock_price_variance/views/product_template_views.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="name">product.template.form.inherit</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='traceability']" position="after">
+                <field name="price_variance_threshold" invisible="1" />
+                <group string="Price Discrepancy Settings" name="price_discrepancy">
+                    <field
+                        name="bypass_price_variance_check"
+                        attrs="{'invisible': [('price_variance_threshold', '=', False)]}"
```
nobuQuartile: @AungKoKoLin1997 
The invisible in this field doesn't work well. It will always be invisible. I don't know why so I would appreciate it if you could teach me.

AungKoKoLin1997: @nobuQuartile  I think this is because of not store. Can you check with this field definition?
`price_variance_threshold = fields.Boolean( related="company_id.price_variance_threshold", readonly=True, store=True, )`

purchase_stock_price_variance/views/product_product_views.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="inherit_id" ref="product.product_normal_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='traceability']" position="after">
+                <field name="price_variance_threshold" invisible="1" />
+                <group string="Price Discrepancy Settings" name="price_discrepancy">
+                    <field
+                        name="bypass_price_variance_check"
+                        attrs="{'invisible': [('price_variance_threshold', '=', False)]}"
+                    />
+                    <field name="price_variance_threshold_percent" />
```
nobuQuartile: @AungKoKoLin1997 
I would like to add % in this view.
I used the code below to refer to it but it didn't work. Is there any good way?

```
　　　　<field
                    name="qty_percentage"
                    attrs="{'invisible': [('advance_payment_method', '!=', 'qty_percentage')]}"
                    widget="percentage"
                />
```

```
　　　　　　　<div>
                                <field name="amount_string" class="oe_inline"/>
                                <span class="o_form_label oe_inline" attrs="{'invisible':[('amount_type','!=','percentage')]}">%</span>
                            </div>
```

nobuQuartile: @AungKoKoLin1997 
Thank you for your comment.
Actually, I tried using the store, but it didn't work well.
2025-02-26_10h50_08
If it seems like it will take more than 15m to Lin-san's investigate, I'm thinking of giving up on hiding it this time or ask Tashiro-san.

purchase_stock_price_variance/views/product_template_views.xml
nobuQuartile: @AungKoKoLin1997
Thank you for your comment.
Actually, I tried using the store, but it didn't work well.
2025-02-26_10h50_08

If it seems like it will take more than 15m to Lin-san's investigate, I'm thinking of giving up on hiding it this time or ask Tashiro-san.

AungKoKoLin1997: There is no issue in your invisible attribute but the problem is price_variance_threshold is not getting the value from company record because the changes of value using related field is not triggered. You should try another way to get the value instead of related field.

purchase_stock_price_variance/views/product_product_views.xml
AungKoKoLin1997: Are you sure using `widget="percentage` is not working?
Screenshot 2025-02-26 090914

nobuQuartile: Sorry. I was misunderstood.
 I found %.

purchase_stock_price_variance/models/product_template.py
```diff
@@ -24,6 +24,15 @@ class ProductTemplate(models.Model):
...
     )
+
+    @api.depends("company_id", "company_id.price_variance_threshold")
+    def _compute_price_variance_threshold(self):
+        for product in self:
+            product.price_variance_threshold = (
+                product.company_id.price_variance_threshold
+                if product.company_id
+                else False
+            )
```
nobuQuartile: This @api.depends does not work.
I've put a breakpoint on line 32,33 and confirmed that this function doesn't work even after changing company_id.price_variance_threshold from the GUI.

The stock_picking code is working fine.

AungKoKoLin1997: If I am not wrong, the values changes in company from related record is not triggered any company related field from another model(eg,product template). May be you can consider using _get_view.

AungKoKoLin1997: @nobuQuartile Please disregard my above comment. Actually, the implementation you did is correct. But I found the root case of why is it not working for related or compute. It is because of you have multiple companies and products are not assigned company automatically (it means company_id field is false). So, it can't get the value 
from current company.

nobuQuartile: Tashiro-san advised me.

```
@api.depends("company_id")
@api.depends_context("company")
```

`company = rec.company_id or self.env.company`
Something like this


Based on that, I created the following code, but it doesn't work.
```
 @api.depends("company_id")
    @api.depends_context("company")
    def _compute_price_variance_threshold(self):
        company = self.company_id or self.env.company
        for rec in self:
            rec.price_variance_threshold = (
                company.price_variance_threshold
                if company
                else False
            )
```
I tried to use @api.depends_context, but it doesn't seem to trigger when the page is loaded. 
While debugging, I put a breakpoint in the company= raw and checked it.
The issue where price_variance_threshold continues to be false has not been resolved

purchase_stock_price_variance/models/product_template.py
```diff
@@ -0,0 +1,39 @@
...
+    bypass_price_variance_check = fields.Boolean(
+        copy=False,
+        tracking=True,
+        help="If enabled, this product will not be checked for price variance between "
+        "the product's standard price and purchase receipt unit price.",
+    )
+    price_variance_threshold = fields.Boolean(
+        compute="_compute_price_variance_threshold",
+        # Set sotre=false to recalculate when page loads.
+        store=False,
```
AungKoKoLin1997: I think you don't need this attribute because default value for compute field is `false`.

purchase_stock_price_variance/models/product_template.py
nobuQuartile: I set price_variance_threshold's store to false so that it recalculates when the page loads
Before:
```
    price_variance_threshold = fields.Boolean(
        compute="_compute_price_variance_threshold",
        store=True,
    )
```
After:
```
    price_variance_threshold = fields.Boolean(
        compute="_compute_price_variance_threshold",
        # Set sotre=false to recalculate when page loads.
        store=False,
    )

```

purchase_stock_price_variance/models/product_template.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    @api.depends("company_id")
+    @api.depends_context("company")
+    def _compute_price_variance_threshold(self):
+        company = self.company_id or self.env.company
```
AungKoKoLin1997: self is record set. So, `self.company_id` should be inside the loop.

nobuQuartile:
@AungKoKoLin1997 @yostashiro 
Please review.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,89 @@
...
+                if (
+                    threshold_percent and percentage_difference > threshold_percent
+                ) or (threshold_amount and amount_difference > threshold_amount):
+                    error_message = (
+                        f"Price variance exceeding a threshold detected for "
+                        f"{product.name}: Received Price = {received_price}, "
+                        f"Proudct Price = {standard_price}."
+                    )
+                    if (
+                        self.env.company.enable_price_variance_error
```
AungKoKoLin1997: There is `enable_price_variance_error ` in picking and why are you using from company record?

purchase_stock_price_variance/models/product_template.py
```diff
@@ -0,0 +1,49 @@
...
+        compute="_compute_enable_price_variance_error",
+    )
+
+    @api.depends("company_id")
+    @api.depends_context("company")
+    def _compute_enable_price_variance_error(self):
+        for rec in self:
+            company = rec.company_id or rec.env.company
+            rec.enable_price_variance_error = (
+                company.enable_price_variance_error if company else False
```
AungKoKoLin1997: How about like this?
```suggestion
        company_id = self.env.company
        for rec in self:
            company = rec.company_id or company_id
            rec.enable_price_variance_error = company.enable_price_variance_error

File: purchase_stock_price_variance/models/stock_picking.py [hunk 28]
```diff
@@ -0,0 +1,89 @@
...
+        return super().write(vals)
+
+    def _action_done(self):
+        global_price_variance_threshold_percent = (
+            self.env.company.price_variance_threshold_percent
+        )
+        global_price_variance_threshold_amount = (
+            self.env.company.price_variance_threshold_amount
+        )
+        for pick in self:
```
AungKoKoLin1997, line 54: Should we skip the loop if `enable_price_variance_error ` is false?
```suggestion
        for pick in self:
            if not pick.enable_price_variance_error:
                 continue 
```

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,89 @@
...
+                    threshold_percent and percentage_difference > threshold_percent
+                ) or (threshold_amount and amount_difference > threshold_amount):
+                    error_message = (
+                        f"Price variance exceeding a threshold detected for "
+                        f"{product.name}: Received Price = {received_price}, "
+                        f"Proudct Price = {standard_price}."
+                    )
+                    if (
+                        self.env.company.enable_price_variance_error
+                        and not pick.bypass_price_variance_check
```
AungKoKoLin1997: This check should be after `for pick in self:`, shouldn't it?

purchase_stock_price_variance/models/stock_picking.py
nobuQuartile: I talked to Tashiro about this yesterday.
Sorry I didn't share it.

The customer was worried that an error would occur and interfere with normal operation. Therefore, we have changed the basic specification to leave logs in the chatter but not errors. The error will occur only when enable_price_variance_error is turned on. If it is turned off, the system will simply log a message in the chatter. The bypass boolean fields have been changed to bypass the error. If the threshold is exceeded regardless of the bypass boolean values, The error log will add to the chatter. In addition, I have changed the bypass boolean fields to be displayed only when enable_price_variance_error is true.

purchase_stock_price_variance/models/product_template.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @api.depends("company_id")
+    @api.depends_context("company")
+    def _compute_enable_price_variance_error(self):
+        company = self.env.company
+        for rec in self:
+            company_id = rec.company_id or company
+            rec.enable_price_variance_error = (
+                company_id.enable_price_variance_error if company_id else False
+            )
```
AungKoKoLin1997: Don't add the condition because `company_id` can't be false in any situation because self.env.company always returns a valid company.

nobuQuartile:
@AungKoKoLin1997 
I spoke with Tashiro, and we decided that my coding skills aren't at the level where I can modify this module. I’m sorry, but could you please handle this instead?

yostashiro:
@AungKoKoLin1997 Please rebase. Is this PR good for review?

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    @api.depends("company_id", "company_id.enable_price_variance_error")
+    def _compute_enable_price_variance_error(self):
+        for picking in self:
+            picking.enable_price_variance_error = (
+                picking.company_id.enable_price_variance_error
+                if picking.company_id
+                else False
+            )
```
yostashiro, line 30:
```suggestion
```

purchase_stock_price_variance/models/product_category.py
```diff
@@ -0,0 +1,25 @@
...
+
+    enable_price_variance_error = fields.Boolean(
+        compute="_compute_enable_price_variance_error",
+    )
+
+    @api.depends_context("company")
+    def _compute_enable_price_variance_error(self):
+        company = self.env.company
+        for rec in self:
+            rec.enable_price_variance_error = company.enable_price_variance_error
```
yostashiro, line 25:
```suggestion
```

purchase_stock_price_variance/views/product_category_views.xml
```diff
@@ -0,0 +1,16 @@
...
+    <record id="view_product_category_form" model="ir.ui.view">
+        <field name="model">product.category</field>
+        <field name="inherit_id" ref="product.product_category_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='parent_id']" position="after">
+                <field name="enable_price_variance_error" invisible="1" />
+                <field
+                    name="bypass_price_variance_check"
+                    attrs="{'invisible': [('enable_price_variance_error', '=', False)]}"
+                />
```
yostashiro, line 12:
```suggestion
                <field name="bypass_price_variance_check" />
```

purchase_stock_price_variance/models/product_template.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    @api.depends("company_id")
+    @api.depends_context("company")
+    def _compute_enable_price_variance_error(self):
+        company = self.env.company
+        for rec in self:
+            rec.enable_price_variance_error = (
+                rec.company_id or company
+            ).enable_price_variance_error
```
yostashiro, line 38:
```suggestion
```

purchase_stock_price_variance/views/product_template_views.xml
```diff
@@ -0,0 +1,30 @@
...
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='traceability']" position="after">
+                <group string="Price Discrepancy Settings" name="price_discrepancy">
+                    <field name="enable_price_variance_error" invisible="1" />
+                    <field
+                        name="bypass_price_variance_check"
+                        attrs="{'invisible': [('enable_price_variance_error', '=', False)]}"
+                    />
```
yostashiro, line 14:
```suggestion
                    <field name="bypass_price_variance_check" />
```

purchase_stock_price_variance/views/stock_picking_views.xml
```diff
@@ -0,0 +1,23 @@
...
+                    string="Price Discrepancy Settings"
+                    name="price_discrepancy"
+                    attrs="{'invisible': [('enable_price_variance_error', '=', False)]}"
+                >
+                    <field
+                        name="bypass_price_variance_check"
+                        attrs="{'readonly': [('state', 'in', ('done','cancel'))]}"
+                    />
+                </group>
+            </xpath>
```
yostashiro, line 20:
```suggestion
            <xpath expr="//group[@name='other_infos']" position="inside">
                <field
                    name="bypass_price_variance_check"
                    attrs="{'readonly': [('state', 'in', ('done','cancel'))]}"
                />
            </xpath>
```

purchase_stock_price_variance/views/product_template_views.xml
```diff
@@ -0,0 +1,30 @@
...
+                        name="price_variance_threshold_percent"
+                        widget="percentage"
+                        class="oe_inline"
+                        style="vertical-align:baseline"
+                    />
+                    <field
+                        name="price_variance_threshold_amount"
+                        class="oe_inline"
+                        style="vertical-align:baseline"
+                    />
```
yostashiro, line 25:
```suggestion
                    <field name="price_variance_threshold_percent" widget="percentage" />
                    <field name="price_variance_threshold_amount" />
```

purchase_stock_price_variance/views/product_template_views.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_product_template_form_inherit" model="ir.ui.view">
+        <field name="name">product.template.form.inherit</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="product.product_template_only_form_view" />
```
yostashiro, line 6:
```suggestion
    <record id="view_template_property_form" model="ir.ui.view">
        <field name="name">product.template.stock.property.form.inherit</field>
        <field name="model">product.template</field>
        <field name="inherit_id" ref="stock.view_template_property_form" />
```

purchase_stock_price_variance/views/product_product_views.xml
```diff
@@ -0,0 +1,30 @@
...
+                    <field
+                        name="price_variance_threshold_amount"
+                        class="oe_inline"
+                        style="vertical-align:baseline"
+                    />
+                </group>
+            </xpath>
+        </field>
+    </record>
+</odoo>
```
yostashiro, line 30:
```suggestion
```
Suppose we don't need this file. Please double check.

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,102 @@
...
+                    continue
+                product = move.product_id
+                threshold_percent = (
+                    product.price_variance_threshold_percent
+                    or global_price_variance_threshold_percent
+                )
+                threshold_amount = (
+                    product.price_variance_threshold_amount
+                    or global_price_variance_threshold_amount
+                )
```
yostashiro, line 68:
```suggestion
                )
                if not threshold_percent and not threshold_amount:
                    continue
```

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,102 @@
...
+        return super().write(vals)
+
+    def _action_done(self):
+        global_price_variance_threshold_percent = (
+            self.env.company.price_variance_threshold_percent
+        )
+        global_price_variance_threshold_amount = (
+            self.env.company.price_variance_threshold_amount
+        )
```
yostashiro, line 53:
```suggestion
        company = self.env.company
        global_threshold_percent = company.price_variance_threshold_percent
        global_threshold_amount = company.price_variance_threshold_amount
```
Not a fan of log variable names in methods. ;)

purchase_stock_price_variance/models/stock_picking.py
```diff
@@ -0,0 +1,102 @@
...
+                    _(
+                        "Price variance exceeding a threshold detected for "
+                        "the following products:\n" + "\n".join(error_messages)
+                    )
+                )
+            if messages:
+                pick.message_post(
+                    body="Price variance exceeding a threshold detected for "
+                    "the following products:<br/>" + "<br/>".join(messages)
+                )
```
yostashiro, line 101:
```suggestion
            def get_message(products, delimiter):
                message = _("Price variance exceeding a threshold detected for the following products:")
                return message + delimiter + delimiter.join(products)

            if pick.company_id.enable_price_variance_error and error_messages:
                raise UserError(get_message(error_messages, "\n"))
            if messages:
                pick.message_post(body=get_message(messages, "<br/>"))
```

purchase_stock_price_variance/readme/CONFIGURE.md
```diff
@@ -0,0 +1,17 @@
...
+    * **Price Variance Threshold Amount**: Default maximum variance for all products.
+4. Go to Inventory > Configuration > Product Categories.
+5. Enable **Bypass Price Variance Check** to skip the error check for the products under this category.
+6. Go to Inventory > Products and open the product.
+7. Click on the Inventory tab and configure the following fields.
+    If the threshold value is set to 0, the threshold will refer to the global value.
+    * **Bypass Price Variance Check**: Enable this to skip the error check for this specific product.
+    * **Price Variance Threshold Percent**: Set the allowable percentage variance.
+    * **Price Variance Threshold Amount**: Define the maximum allowable price variance.
+8. Assign the internal user to the "Bypass Price Variance Check" group.
```
yostashiro, line 16: I think the name of the group is misleading. "Manage Price Variance Check" is better.

yostashiro (changes requested):

AungKoKoLin1997:
This PR was already proposed in OCA repo. Closing this PR.
https://github.com/OCA/stock-logistics-workflow/pull/1893

PR #113, opened 2025-04-21
[5300][IMP] stock_lot_suffix: udpate suffix when analytic_distribution is updated

AungKoKoLin1997:
[5300](https://www.quartile.co/web#id=5300&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)


stock_lot_suffix/models/stock_lot.py
```diff
@@ -22,3 +22,14 @@ def name_get(self):
...
+    def write(self, vals):
+        res = super().write(vals)
+        if "analytic_distribution" not in vals:
+            return res
+        for rec in self:
+            rec.lot_suffix = False
+            analytic_account = rec.analytic_account_ids.filtered(lambda x: x.lot_suffix)
+            if analytic_account:
+                rec.lot_suffix = analytic_account[0].lot_suffix
+        return res
```
nobuQuartile: Is it wrong to use the compute attribute?

nobuQuartile (approved):
Functional Review

nobuQuartile:
I think you can delete this section.
https://github.com/qrtl/axls-custom/blob/16.0/stock_lot_suffix/models/stock_move_line.py#L23C13-L28C68

stock_lot_suffix/models/stock_lot.py
AungKoKoLin1997: It makes sense.

AungKoKoLin1997:
The latest changes depends on https://github.com/qrtl/axls-oca/pull/187.

nobuQuartile (approved):
LGTM
Functional test and code review

stock_lot_suffix/models/stock_lot.py
```diff
@@ -22,3 +22,12 @@ def name_get(self):
...
         return new_name_list
+
+    @api.depends("analytic_account_ids")
+    def _compute_lot_suffix(self):
+        for rec in self:
+            rec.lot_suffix = False
+            # We assume that there is only one analytic account with lot_suffix if any.
+            analytic_account = rec.analytic_account_ids.filtered(lambda x: x.lot_suffix)
+            if analytic_account:
+                rec.lot_suffix = analytic_account[0].lot_suffix
```
yostashiro: To be slightly shorter.
```suggestion
            # We assume that there is only one analytic account with lot_suffix if any.
            aa = rec.analytic_account_ids.filtered(lambda x: x.lot_suffix, limit=1)
            rec.lot_suffix = aa.lot_suffix if aa else False
```

stock_lot_suffix/models/stock_move_line.py
yostashiro, line 1: Would it be more appropriate to extend the `stock.move` method, instead?

AungKoKoLin1997: I prefer `stock.move.line` because it is for each lot of move line.

stock_lot_suffix/models/stock_lot.py
AungKoKoLin1997: I don't think there is `limit` attribute for `filtered`.
Even we use like `rec.analytic_account_ids.filtered(lambda x: x.lot_suffix)[0]`, will throw error if there is no accounts.

yostashiro: Then, `rec.analytic_account_ids.filtered(lambda x: x.lot_suffix)[:1]`?

stock_lot_suffix/models/stock_move_line.py
yostashiro: My point is to do `write()` on the multiple `stock.move.line` records at once for each move, which should be more efficient. But it may not make a big difference after all.

AungKoKoLin1997: I see. I think it will be better.

stock_lot_suffix/models/stock_move.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    def _action_done(self, cancel_backorder=False):
+        res = super()._action_done(cancel_backorder=cancel_backorder)
+        for move in self:
+            if not move.lot_ids:
```
yostashiro: I think the original logic may have had an issue (looks like non-purchase related moves would wipe out the `channel_category` value of the lot). @nobuQuartile Can you please help confirm this point?
```suggestion
            if not move.purchase_line_id or not move.lot_ids:
```

nobuQuartile: @yostashiro 
You are correct. I have verified in my development environment that non-purchase-related moves set the channel_category value of the lot to false.

yostashiro (approved):
Code review.

nobuQuartile (approved):
Code review, Functional test, LGTM

kanda999 (approved):

PR #122, opened 2025-07-22
[5540][ADD] analytic_distribution_filter

AungKoKoLin1997:
[5540](https://www.quartile.co/web#id=5540&cids=3&menu_id=517&model=project.task&view_type=form)


AungKoKoLin1997:
@kanda999 Can you please check the behavior? 
Currently it is provided the reverse auto assign for related accounts.(it means the reverse account is not assigned in related account automatically if we assign that account as a related account for some account).
Seem like it brings complexity in code if we want to do it because it can cause recursion and we need to block it and consider on all CURD methods(create, write, unlink).
Please let me know if you have an clean approach for this.

AungKoKoLin1997:
> @kanda999 Can you please check the behavior? Currently it is provided the reverse auto assign for related accounts.(it means the reverse account is not assigned in related account automatically if we assign that account as a related account for some account). Seem like it brings complexity in code if we want to do it because it can cause recursion and we need to block it and consider on all CURD methods(create, write, unlink). Please let me know if you have an clean approach for this.

@kanda999 I added new commit for this. Please check it out.

analytic_distribution_filter/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,3 @@
+This module introduces a filter for the analytic distribution,
+configurable per model via an “Apply Analytic Distribution Filter”
```
kanda999:
```suggestion
configurable per model via an "Apply Analytic Distribution Filter"
```

analytic_distribution_filter/models/analytic_account_line.py
```diff
@@ -0,0 +1,83 @@
...
+        return res
+
+    def write(self, vals):
+        for record in self:
+            old_ids = record.account_ids.ids
+            super().write(vals)
+            record._update_reciprocal_relations(old_account_ids=old_ids)
+        return True
+
+    def unlink(self):
```
yostashiro, line 73: Use @api.ondelete() instead?

analytic_distribution_filter/models/analytic_account_line.py
```diff
@@ -0,0 +1,83 @@
...
+                )
+                if reciprocal:
+                    if line.account_id.id not in reciprocal.account_ids.ids:
+                        reciprocal.account_ids = [(4, line.account_id.id)]
+                else:
+                    self.create(
+                        {
+                            "account_id": account_id,
+                            "plan_id": line.account_id.plan_id.id,
+                            "account_ids": [(6, 0, [line.account_id.id])],
```
yostashiro: Use Command.set().

analytic_distribution_filter/models/ir_model.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrModel(models.Model):
+    _inherit = "ir.model"
+
+    apply_analytic_distribution_filter = fields.Boolean(
```
yostashiro: Didn't quite get the design rationale. Do we have cases where we want to apply a filter on one model and not on another?

analytic_distribution_filter/models/analytic_account.py
```diff
@@ -0,0 +1,20 @@
...
+    related_account_ids = fields.Many2many(
+        "account.analytic.account",
+        string="Related accounts (all plans)",
+        compute="_compute_related_accounts",
+    )
+
+    @api.depends("account_line_ids.account_ids")
+    def _compute_related_accounts(self):
+        for rec in self:
+            rec.related_account_ids = rec.account_line_ids.mapped("account_ids")
```
yostashiro: Can assume that all analytic accounts of a plan are "related" if no account of the plan is linked?

yostashiro (changes requested):
Partial review.

analytic_distribution_filter/models/analytic_account_line.py
AungKoKoLin1997: I checked this decorator and this is for client-server errors when unlinking records.
So, I don't want to use this decorator for our case.

AungKoKoLin1997:
Closing this PR and will propose in axls-oca after OCA PR has approved.

PR #126, opened 2025-10-30
[5799][ADD] ir_model_access_adj

nobuQuartile:
[5799](https://www.quartile.co/web?db=qrtl&token=z9OpY14OSBGGYn2qv0Ha#id=5799&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)


yostashiro (changes requested):
Can we instead update product_lot_sequence?

- Update `_create_lot_sequence()` to assign `"code": "product_lot_sequence"`
- Create a new group "Manage product lot sequence"
- Create an ACL record (same as what you suggest in this PR except for the group) and a record rule for the new group

The record rule should be something like:

```.xml
<record id="rule_ir_sequence_product_lot_sequence_only" model="ir.rule">
    <field name="name">Product Lot Sequence Manager: product lot sequences only</field>
    <field name="model_id" ref="base.model_ir_sequence"/>
    <field name="groups" eval="[Command.link(ref('product_lot_sequence.group_product_lot_sequence_manager'))]"/>
    <field name="domain_force">
        [
            ("code", "=", "product_lot_sequence"),
            ("company_id", "in", company_ids + [False]),
        ]
    </field>
</record>
```

Once the module is updated, "Manage product lot sequence" should be assigned to the users who need it. We may help update `code` of the existing sequences as necessary.

AungKoKoLin1997:
> The record rule should be something like:
> 
> ```
> <record id="rule_ir_sequence_product_lot_sequence_only" model="ir.rule">
>     <field name="name">Product Lot Sequence Manager: product lot sequences only</field>
>     <field name="model_id" ref="base.model_ir_sequence"/>
>     <field name="groups" eval="[Command.link(ref('product_lot_sequence.group_product_lot_sequence_manager'))]"/>
>     <field name="domain_force">
>         [
>             ("code", "=", "product_lot_sequence"),
>             ("company_id", "in", company_ids + [False]),
>         ]
>     </field>
> </record>
> ```

@yostashiro 
I checked the design, and it turns out to be too risky to implement this rule.
With the current configuration, the document sequence generation is blocked whenever a user in this new group triggers it, since the rule only allows access to lot sequences for that group.

We could theoretically add another rule to grant access to other sequences and fix this issue, but doing so would not be ideal — it might override or interfere with existing access rules defined by other modules.

```
<record id="rule_ir_sequence_read_all" model="ir.rule">
    <field name="name">Read access to all sequences</field>
    <field name="model_id" ref="base.model_ir_sequence"/>
    <field name="groups" eval="[Command.link(ref('product_lot_sequence.group_product_lot_sequence_manager'))]"/>
    <field name="perm_read" eval="1"/>
    <field name="perm_write" eval="0"/>
    <field name="perm_create" eval="0"/>
    <field name="perm_unlink" eval="0"/>
    <field name="domain_force">[(1, "=", 1)]</field>
</record>
```

nobuQuartile:
We will follow up on the PR.

[5799][IMP] product_lot_sequence: reflect oca changes
https://github.com/qrtl/axls-oca/pull/208

PR #130, opened 2025-11-26
[5661][ADD] purchase_deposit_preserve_amount

AungKoKoLin1997:
[5661](https://www.quartile.co/web#id=5661&cids=3&menu_id=505&action=1457&model=project.task&view_type=form)

With current code, the amount on the vendor bill is worked as expected behavior. But I need to work on stock valuation adjustment.


AungKoKoLin1997:
@kanda999 I updated the module as per our yesterday discussion.
Could you please review and let me know if you have any update about refund process.

purchase_deposit_preserve_amount/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+    _inherit = "account.move.line"
+
+    def _generate_price_difference_vals(self, layers):
+        self.ensure_one()
+        is_deposit_move = self.move_id.line_ids.filtered(
+            lambda line: line.purchase_line_id.is_deposit
+        )
+        if not is_deposit_move:
+            return super()._generate_price_difference_vals(layers)
+        self = self.with_context(aml=self)
```
yostashiro: Please make the context key more explicit.

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2025 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    is_deposit = fields.Boolean(string="Is Deposit Move", readonly=True)
```
yostashiro: Shouldn't this be made an computed field instead?

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def _adjust_journal_item_balances_for_deposit(self):
+        for rec in self:
+            deposit_lines = rec.line_ids.filtered(
+                lambda line: line.display_type == "product"
+                and line.purchase_line_id.is_deposit
+            )
```
yostashiro, line 30: To slightly improve the clarity and performance.
```suggestion
            )
            if not deposit_lines:
                continue
```

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+            amount_diff = 0.0
+            for line in deposit_lines:
+                balance = line.purchase_line_id.invoice_lines.filtered(
+                    lambda l: l.move_id.state == "posted" and l.move_id.is_deposit
+                ).balance
+                amount_diff += balance + line.balance
+                line.with_context(skip_deposit_adjustment=True).balance = -1 * balance
```
yostashiro, line 41: A minor point. I prefer early exits to reduce computation.
```suggestion
                line.with_context(skip_deposit_adjustment=True).balance = -1 * balance
            if not amount_diff:
                continue
```

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+            product_lines = rec.line_ids.filtered(
+                lambda line: line.display_type == "product"
+                and line.purchase_line_id
+                and not line.purchase_line_id.is_deposit
+            )
+            if not product_lines or not amount_diff:
+                continue
+            currency = rec.currency_id or rec.company_id.currency_id
```
yostashiro: Is there actually any case where rec.currency_id is false?

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,73 @@
...
+            line_count = len(product_lines)
+            # 2) We want: sum(adjustments) == amount_diff
+            #    First N-1 lines get rounded equal shares,
+            #    last line gets the exact remainder.
+            equal_share = amount_diff / line_count
```
yostashiro: I feel that this equal distribution is strange. I think it should be weighted by line.balance.

yostashiro (changes requested):
Partial review.

purchase_deposit_preserve_amount/models/account_move.py
AungKoKoLin1997: I think we shouldn't because it is hard to fill by depending on purchase order since we can have both deposit bill and normal vendor bill. So, the best way is filling the value when the deposit bill is created.

AungKoKoLin1997: Done!

kanda999 (approved):
Functional review

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,83 @@
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        moves = super().create(vals_list)
+        moves._moves_needing_deposit_adjustment()._adjust_journal_item_balances_for_deposit()
+        return moves
+
+    def write(self, vals):
+        if self.env.context.get("skip_deposit_adjustment"):
+            return super().write(vals)
+        res = super().write(vals)
```
yostashiro:
```suggestion
        res = super().write(vals)
        if self.env.context.get("skip_deposit_adjustment"):
            return res
```

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,83 @@
...
+                lambda line: line.display_type == "product"
+                and line.purchase_line_id
+                and not line.purchase_line_id.is_deposit
+            )
+            if not product_lines:
+                continue
+            currency = rec.currency_id
+            line_count = len(product_lines)
+            total_balance = sum(product_lines.mapped("balance"))
+            remaining = amount_diff
```
yostashiro: A minor point but I opt for reducing variables if there is no logical necessity and they don't contribute to reducing the code volume. For example, we can remove `remaining = amount_diff` and name `amount_diff` as `amount_diff_bal` from the start.

purchase_deposit_preserve_amount/models/account_move.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def _compute_is_deposit(self):
+        for rec in self:
+            rec.is_deposit = any(
+                rec.invoice_line_ids.filtered(
+                    lambda line: line.purchase_line_id.is_deposit and line.quantity > 0
+                )
+            )
+
+    def _adjust_journal_item_balances_for_deposit(self):
```
yostashiro, line 20: A comment block explaining the idea would help.

purchase_deposit_preserve_amount/models/account_move_line.py
```diff
@@ -0,0 +1,18 @@
...
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _generate_price_difference_vals(self, layers):
+        self.ensure_one()
+        is_deposit_move = self.move_id.line_ids.filtered(
+            lambda line: line.purchase_line_id.is_deposit
+        )
+        if not is_deposit_move:
```
yostashiro:
```suggestion
        if not self.move_id.line_ids.filtered(
            lambda line: line.purchase_line_id.is_deposit
        ):
```

yostashiro (approved):

AungKoKoLin1997:
I reopen this PR.

AungKoKoLin1997:
I added new commit depends on odoo-review.

PR #131, opened 2025-12-13
[5847][IMP] account_move_obc_csv

kanda999:
[5847](https://www.quartile.co/web#id=5847&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)

### DIFF
--- a/account_move_obc_csv/report/report_account_move_obc_csv.py
+++ b/account_move_obc_csv/report/report_account_move_obc_csv.py
@@ -215,6 +215,10 @@ class AccountMoveObcCsv(models.AbstractModel):
 
     def generate_csv_report(self, writer, data, records):
         self._check_records(records)
+        # NOTE: Some records may be intentionally excluded from the CSV output
+        # but they should still be flagged as exported
+        # to prevent duplicate processing later.
+        records.is_exported = True
         # Exclude journal entries related to purchase receipts from export.
         records = records.filtered(
             lambda m: not self._is_purchase_receipt_valuation_move(m)
@@ -254,7 +258,6 @@ class AccountMoveObcCsv(models.AbstractModel):
                 vals_dict[1]["GL0010000"] = "*"
             for _k, v in sorted(vals_dict.items()):
                 writer.writerow(v)
-            rec.is_exported = True
 
     def csv_report_options(self):
         res = super().csv_report_options()
--- a/account_move_obc_csv/report/report_account_move_obc_csv.py
+++ b/account_move_obc_csv/report/report_account_move_obc_csv.py
@@ -85,13 +85,47 @@ class AccountMoveObcCsv(models.AbstractModel):
         production = stock_move.production_id or stock_move.raw_material_production_id
         return production.subcontractor_id
 
+    def _is_purchase_receipt_valuation_move(self, move):
+        """
+        Identify account moves for purchase receipts (incoming pickings) that should
+        NOT be exported to OBC.
+        """
+        stock_move = move.stock_move_id
+        if not stock_move or not stock_move.picking_id:
+            return False
+        picking_type = stock_move.picking_id.picking_type_id
+        if not picking_type or picking_type.code != "incoming":
+            return False
+        if not stock_move.purchase_line_id:
+            return False
+        return True
+
     def _update_vals(self, vals, line, move_analytic_accounts, drcr):
-        account_code = line.account_id.code
+        account = line.account_id
+        account_code = account.code
         subaccount_code = ""
         if "." in account_code:
             # maxsplit=1 - we assume that an account code should contain only one
             # period (".") at most.
             account_code, subaccount_code = account_code.split(".", 1)
+        # Vendor bill journal export:
+        # - If the line uses the purchase accrual (Goods Received Not Invoiced) account,
+        #   output it as "material purchase" by mapping it to the product category's
+        #   stock valuation account.
+        move = line.move_id
+        categ = line.product_id.categ_id if line.product_id else False
+        if (
+            move.move_type in ("in_invoice", "in_refund")
+            and categ
+            and categ.property_stock_account_input_categ_id
+            and account == categ.property_stock_account_input_categ_id
+            and categ.property_stock_valuation_account_id
+        ):
+            account_code = categ.property_stock_valuation_account_id.code
+            if "." in account_code:
+                account_code, subaccount_code = account_code.split(".", 1)
+            else:
+                subaccount_code = ""
         # We assume that there should be only one project/department per journal
         # item if any.
         project = line.analytic_line_ids.filtered(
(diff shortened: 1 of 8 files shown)


account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,13 +85,47 @@ def _get_partner(self, line):
...
         if "." in account_code:
             # maxsplit=1 - we assume that an account code should contain only one
             # period (".") at most.
             account_code, subaccount_code = account_code.split(".", 1)
+        # Vendor bill journal export:
+        # - If the line uses the purchase accrual (Goods Received Not Invoiced) account,
+        #   output it as "material purchase" by mapping it to the product category's
+        #   stock valuation account.
+        move = line.move_id
+        categ = line.product_id.categ_id if line.product_id else False
```
yostashiro: We can shorten it.
```suggestion
        categ = line.product_id.categ_id
```

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,13 +85,47 @@ def _get_partner(self, line):
...
+        if (
+            move.move_type in ("in_invoice", "in_refund")
+            and categ
+            and categ.property_stock_account_input_categ_id
+            and account == categ.property_stock_account_input_categ_id
+            and categ.property_stock_valuation_account_id
+        ):
+            account_code = categ.property_stock_valuation_account_id.code
```
yostashiro: I think this part should come at the beginning of the method, to be followed by the `account_code.split()` part.
```suggestion
        categ = line.product_id.categ_id
        if not line.is_landed_cost_line and line.move_id.is_purchase_document(include_receipts=True):
            account_input = categ.property_stock_account_input_categ_id
            account_valuation = categ.property_stock_valuation_account_id
            if line. account_valuation and line.account_id == account_input:
                account_code = account_valuation
```

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -119,15 +153,8 @@ def _update_vals(self, vals, line, move_analytic_accounts, drcr):
             department = move_analytic_accounts.filtered(
                 lambda x: x.plan_id.plan_type == "department"
             )[:1]
-        # Records of purchase interim account should not be passed to OBC as taxable,
-        # due to conceptual discrepancies between real-time inventory accounting and
-        # 3-part method.
         tax = self.env["account.tax"]
```
yostashiro: Remove this line as well.
```suggestion
```

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,13 +85,47 @@ def _get_partner(self, line):
         production = stock_move.production_id or stock_move.raw_material_production_id
         return production.subcontractor_id
 
+    def _is_purchase_receipt_valuation_move(self, move):
+        """
+        Identify account moves for purchase receipts (incoming pickings) that should
```
yostashiro:
```suggestion
        """Identify account moves for purchase receipts (incoming pickings) that should
```

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -188,6 +215,10 @@ def _get_report_vals_dict(self, record):
 
     def generate_csv_report(self, writer, data, records):
         self._check_records(records)
+        # Exclude journal entries related to purchase receipts from export.
+        records = records.filtered(
+            lambda m: not self._is_purchase_receipt_valuation_move(m)
+        )
```
yostashiro: Shouldn't we show error by adjusting _check_records() instead when user selects receipt moves?

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,13 +85,47 @@ def _get_partner(self, line):
...
+        """
+        stock_move = move.stock_move_id
+        if not stock_move or not stock_move.picking_id:
+            return False
+        picking_type = stock_move.picking_id.picking_type_id
+        if not picking_type or picking_type.code != "incoming":
+            return False
+        if not stock_move.purchase_line_id:
+            return False
+        return True
```
yostashiro: @kanda999 It wasn't clear to me why you checked `sm.picking_code != "incoming"`.

```suggestion
        return True if move.stock_move_id.purchase_line_id else False
```

yostashiro (changes requested):

account_move_obc_csv/report/report_account_move_obc_csv.py
kanda999: @yostashiro 
I decided to keep this part as it is.
The reason is that I assumed users would not want to leave any journal entries with is_exported = False.

The flow I have in mind is as follows:
- Go to Accounting > Journal Entries.
- Select all records where is_exported = False.
- Export them.

I thought it would not be good design to block this flow. 
Also, since the filter I added in this PR is not used in this flow, I removed it in the latest commit.

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,9 +85,27 @@ def _get_partner(self, line):
...
+        #   output it as "material purchase" by mapping it to the product category's
+        #   stock valuation account.
+        categ = line.product_id.categ_id
+        if not line.is_landed_costs_line and line.move_id.is_purchase_document(
+            include_receipts=True
+        ):
+            account_input = categ.property_stock_account_input_categ_id
+            account_valuation = categ.property_stock_valuation_account_id
+            if account_valuation and line.account_id == account_input:
+                account_code = account_valuation.code
```
yostashiro: Looking at the structure again, I think we should probably have a dedicated helper method for this conversion, and call it in _get_report_vals_dict() where move_lines are looped.

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -177,17 +193,29 @@ def _get_report_vals_dict(self, record):
                 )
             vals["GL0011001"] = remarks
             if line.debit:
-                vals = self._update_vals(vals, line, move_analytic_accounts, "dr")
+                vals = self._update_vals(
+                    vals, line, move_analytic_accounts, "dr", account_code
+                )
```
yostashiro, line 192: I think we can assign account_code and subaccount_code to vals at the beginning of the loop, and then we don't have to change the method args.

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -177,17 +192,29 @@ def _get_report_vals_dict(self, record):
...
     def generate_csv_report(self, writer, data, records):
         self._check_records(records)
+        # NOTE: Some records may be intentionally excluded from the CSV output
+        # but they should still be flagged as exported
+        # to prevent duplicate processing later.
+        records.is_exported = True
+        # Exclude journal entries related to purchase receipts from export.
+        records = records.filtered(
+            lambda m: not self._is_purchase_receipt_valuation_move(m)
+        )
```
yostashiro:
```suggestion
        records = records.filtered(lambda m: not m.stock_move_id.purchase_line_id)
```

account_move_obc_csv/report/report_account_move_obc_csv.py
```diff
@@ -85,8 +85,25 @@ def _get_partner(self, line):
...
+        """Get the account code to use for OBC export.
+        For vendor bill lines using the purchase accrual (Goods Received Not Invoiced)
+        account, return the stock valuation account instead to represent it as
+        "material purchase" in the OBC system."""
         account_code = line.account_id.code
+        categ = line.product_id.categ_id
+        if (
+            categ
+            and not line.is_landed_costs_line
+            and line.move_id.is_purchase_document(include_receipts=True)
```
yostashiro, line 98: It's actually better to check this outside the line loop inside _get_report_vals_dict() and update the context, and then skip _get_account_for_export() when the context is true.

yostashiro (approved):

AungKoKoLin1997 (approved):
Code Review: LGTM

PR #133, opened 2025-12-17
[6206][IMP] inventory_reporting

nobuQuartile:
[6206](https://www.quartile.co/web?db=qrtl&token=z9OpY14OSBGGYn2qv0Ha#id=6206&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)

### DIFF
--- a/inventory_reporting/reports/inventory_report.py
+++ b/inventory_reporting/reports/inventory_report.py
@@ -130,46 +130,27 @@ class InventoryReportXlsx(models.AbstractModel):
             {
                 "name": _("Component Flush"),
                 "filter": [
-                    "|",
-                    "&",
                     ("stock_move_id.location_dest_id.usage", "=", "production"),
                     (
                         "stock_move_id.location_id.is_subcontracting_location",
                         "=",
                         False,
                     ),
-                    "&",
-                    ("stock_move_id.location_id.usage", "=", "production"),
-                    (
-                        "stock_move_id.location_dest_id.is_subcontracting_location",
-                        "=",
-                        False,
-                    ),
-                    ("stock_move_id.picking_code", "in", ("internal", "outgoing")),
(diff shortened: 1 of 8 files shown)


yostashiro (approved):
Code review. LGTM.

nobuQuartile:
Since only stock_moves going to production location were being used as outbound movements (Component Flush) for inventory reporting, I renamed Customer Return to Component Return and limited the filter to only include returns of those Component Flush movements' return.

inventory_reporting/reports/inventory_report.py
```diff
@@ -150,6 +150,28 @@ def get_storable_categories(self, wizard):
...
+                    ),
+                    "&",
+                    ("stock_move_id.location_id.usage", "=", "production"),
+                    (
+                        "stock_move_id.location_dest_id.is_subcontracting_location",
+                        "=",
+                        False,
+                    ),
+                    ("stock_move_id.picking_code", "=", "incoming"),
+                    ("stock_move_id.origin_returned_move_id", "!=", False),
```
yostashiro: @nobuQuartile I'm now not very clear what stock moves we are trying to capture here. Can you please list such records (from the copy database) in the task?

Could we remove the leaf `("stock_move_id.picking_code", "=", "incoming"),`, as I think this condition is probably not reliable? Please check this against the real data in the copy database.

We may review and revisit the conditions of "Component Flush" as well.

nobuQuartile: > Could we remove the leaf ("stock_move_id.picking_code", "=", "incoming"),, as I think this condition is probably not reliable? Please check this against the real data in the copy database.

You are right. I removed it.

nobuQuartile:
We decided to follow up on this change by creating another module in the PR below.
https://github.com/qrtl/axls-custom/pull/134

I will close this PR.

PR #134, opened 2026-01-19
[6206][ADD] stock_valuation_layer_reporting

nobuQuartile:
[6206](https://www.quartile.co/web?db=qrtl&token=z9OpY14OSBGGYn2qv0Ha#id=6206&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)


nobuQuartile:
Why the action-based SVL report_category update (wizard) approach
- SVL is read‑only by design; enabling inline edit triggers domain validation errors.
- An action + wizard lets users safely update report_category in bulk without making the SVL view editable.

nobuQuartile:
Ready to review
Please review the last commit.

stock_valuation_layer_reporting/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_svl_report_wizard_account_manager,svl.report.wizard,model_svl_report_wizard,account.group_account_manager,1,1,1,1
+access_svl_report_category_stock_manager,svl.report.category,model_svl_report_category,account.group_account_manager,1,1,1,1
```
kanda999:
```suggestion
access_svl_report_category_stock_manager,svl.report.category,model_svl_report_category,account.group_stock_manager,1,1,1,1
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,384 @@
...
+            25,  # Product Cost Method
+            30,  # SVL's Total Inventory Value
+            30,  # Analytic Distribution
+        ]
+        for col, width in enumerate(column_widths):
+            ws.set_column(col, col, width)
+        for col, header in enumerate(headers):
+            ws.write(0, col, header)
+
+    def generate_storable_report(self, workbook, wizard):
```
kanda999: This method is almost identical to `generate_consumable_report`.

Can't we combine them?

kanda999 (changes requested):

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,37 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        for vals, record in zip(vals_list, records):
+            if not vals.get("report_category"):
+                record.report_category = record._get_default_report_category()
+        return records
+
+    def _get_default_report_category(self):
```
yostashiro: Please move this above create().

stock_valuation_layer_reporting/wizards/svl_report_category_wizard.py
```diff
@@ -0,0 +1,23 @@
...
+    _name = "svl.report.category.wizard"
+    _description = "SVL Report Category Wizard"
+
+    report_category_id = fields.Many2one(
+        "svl.report.category",
+        string="Report Category",
+        required=True,
+    )
+
+    def action_apply(self):
```
yostashiro, line 17: The current design is a bit risky in that it makes you lose track of the original value once the action is run. Any countermeasure on this point?

yostashiro (commented):
Partial review.

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: Done.
Could you look at my latest commit?

stock_valuation_layer_reporting/security/ir.model.access.csv
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
...
+    _inherit = "stock.valuation.layer"
+
+    default_report_category = fields.Many2one(
+        "svl.report.category",
+    )
+
+    report_category = fields.Many2one(
+        "svl.report.category",
+    )
+
```
AungKoKoLin1997, line 31:
```suggestion
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
...
+from odoo import api, fields, models
+
+
+class StockValuationLayer(models.Model):
+    _inherit = "stock.valuation.layer"
+
+    default_report_category = fields.Many2one(
+        "svl.report.category",
+    )
+
```
AungKoKoLin1997:
```suggestion
```

stock_valuation_layer_reporting/wizards/svl_report_category_wizard.py
nobuQuartile: I have added the default_report_category field, the report_category_changed field, and the report_category_changed filter to track the original value.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
...
+    @api.depends("default_report_category", "report_category")
+    def _compute_report_category_changed(self):
+        for record in self:
+            record.report_category_changed = (
+                record.default_report_category != record.report_category
+            )
+
+    def _get_report_category_for_record(self, categories, other_category):
```
AungKoKoLin1997: I don't quite get this logic. Shouldn't we use first match record if there is matched records.
Then, I think `other_category` parameters is not needed to use in `apply_report_category_defaults()` and pass to this method since this is only for current method. You can just define in current method.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        other_category = categories.filtered("is_other")[:1]
+        for record in self:
+            category_value = record._get_report_category_for_record(
+                categories, other_category
+            )
+            if not category_value:
+                continue
+            vals = {"report_category": category_value.id}
+            if not record.default_report_category:
+                vals["default_report_category"] = category_value.id
```
AungKoKoLin1997: Does it mean we don't need to update `default_report_category` for the old records that have been added even we create new categories records and the value from` _get_report_category_for_record` change but need to update `report_category`? Seem like strange logic for me because these two fields will be updated same value at the first time when there are no values but report_category will only be updated when we added new categories and run `Recompute Report Categories`.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2023 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class StockValuationLayer(models.Model):
+    _inherit = "stock.valuation.layer"
+
+    default_report_category = fields.Many2one(
```
AungKoKoLin1997: Can this field be non stored compute field and remove assigning from `apply_report_category_defaults`?

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        for category in categories:
+            if category.is_other:
+                continue
+            if self.filtered_domain(category._get_domain()):
+                matches.append(category)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def apply_report_category_defaults(self):
+        categories = self.env["svl.report.category"].search(
+            [("active", "=", True)], order="sequence,id"
```
AungKoKoLin1997:
```suggestion
            [], order="sequence,id"
```

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @api.constrains("domain")
+    def _check_domain(self):
+        for record in self:
+            if not record.domain:
+                continue
+            try:
+                safe_eval(record.domain)
+            except Exception as exc:
+                raise UserError(
```
AungKoKoLin1997: Better to use `ValidationError` for constrains method.

AungKoKoLin1997 (commented):
Partial Review.

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,79 @@
...
+                    name="default_report_category"
+                    groups="base.group_system"
+                    optional="hide"
+                    readonly="0"
+                />
+            </field>
+        </field>
+    </record>
+
+    <record
```
AungKoKoLin1997: Does this server action intent to run often or just one time to assign for existing records?

nobuQuartile: To assign report_category to existing records.
If we update the report_category domain, we can bulk-update  SVL's report_category with this action.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: > I don't quite get this logic. Shouldn't we use first match record if there is matched records.

If there are records that match multiple report_categories, we also intend to categorize them as Other.
This is because Other is meant to indicate records that require manual review/correction.
In the end, when the data is shown in Excel, displaying the records as Other in the report allows us to manually change the report category properly.

> Then, I think other_category parameters is not needed to use in apply_report_category_defaults() and pass to this method since this is only for current method. 

I agree with this point. I will update the code.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: > I agree with this point. I will update the code.

Done

stock_valuation_layer_reporting/models/svl_report_category.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: We decided to remove default_report_category.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,48 @@
...
+        other_category = categories.filtered("is_other")[:1]
+        matches = []
+        for category in categories:
+            if category.is_other:
+                continue
+            if self.filtered_domain(category._get_domain()):
+                matches.append(category)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def apply_report_category_defaults(self):
```
AungKoKoLin1997: Shouldn't we add default method to `report_category ` and hooks instead of using server action, create() and custom method?

nobuQuartile: Thank you for your advice!

> hooks

With post_init_hook, the report category domains haven’t been configured yet, so nothing will be assigned.
Also, recalculating report categories on every upgrade is not appropriate, because it would overwrite any manually adjusted values.

> default method 

I’ll try rewriting the create() logic using a default instead.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,48 @@
...
+            category_value = record._get_report_category_for_record(categories)
+            record.report_category_changed = record.report_category != category_value
+
+    def _get_report_category_for_record(self, categories):
+        self.ensure_one()
+        other_category = categories.filtered("is_other")[:1]
+        matches = []
+        for category in categories:
+            if category.is_other:
+                continue
```
AungKoKoLin1997:
```suggestion
        for category in categories.filtered(lambda c: not c.is_other):
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,48 @@
...
+        "svl.report.category",
+    )
+    report_category_changed = fields.Boolean(
+        compute="_compute_report_category_changed",
+        store=True,
+    )
+
+    @api.depends("report_category")
+    def _compute_report_category_changed(self):
+        categories = self.env["svl.report.category"].search([], order="sequence,id")
```
AungKoKoLin1997: I think you don't need `order` since it was already defined in models.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,48 @@
...
+        matches = []
+        for category in categories:
+            if category.is_other:
+                continue
+            if self.filtered_domain(category._get_domain()):
+                matches.append(category)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def apply_report_category_defaults(self):
+        categories = self.env["svl.report.category"].search([], order="sequence,id")
```
AungKoKoLin1997: Same as above

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,334 @@
...
+        if report_type == "storable":
+            base_domain = expression.AND(
+                [base_domain, [("product_id.detailed_type", "=", "product")]]
+            )
+        else:
+            base_domain = expression.AND(
+                [base_domain, [("product_id.detailed_type", "!=", "product")]]
+            )
+        categories = self.get_inventory_operation_categories(
+            display_type=report_type, include_other=True
```
AungKoKoLin1997:
```suggestion
            display_type=report_type
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,334 @@
...
+        categories = self._get_report_categories().filtered(
+            lambda c: c.display_type in (display_type, "both")
+        )
+        if not include_other:
+            categories = categories.filtered(lambda c: not c.is_other)
+        return [(category.id, category.name) for category in categories]
+
+    def _get_report_categories(self):
+        return self.env["svl.report.category"].search(
+            [("active", "=", True)], order="sequence,id"
```
AungKoKoLin1997:
```suggestion
   
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,334 @@
...
+        self, display_type="storable", include_other=True
+    ):
+        categories = self._get_report_categories().filtered(
+            lambda c: c.display_type in (display_type, "both")
+        )
+        if not include_other:
+            categories = categories.filtered(lambda c: not c.is_other)
+        return [(category.id, category.name) for category in categories]
+
+    def _get_report_categories(self):
```
AungKoKoLin1997: Not sure it is worth to have separate method since it is called from only one method.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,334 @@
...
+                [base_domain, [("product_id.detailed_type", "=", "product")]]
+            )
+        else:
+            base_domain = expression.AND(
+                [base_domain, [("product_id.detailed_type", "!=", "product")]]
+            )
+        categories = self.get_inventory_operation_categories(
+            display_type=report_type, include_other=True
+        )
+        self._setup_summary_sheet(workbook, categories, report_type)
```
AungKoKoLin1997: I think we can save mutiple line and readability in `_setup_summary_sheet` if we just pass by mapping the name for `categories`.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def get_inventory_operation_categories(
+        self, display_type="storable", include_other=True
+    ):
+        categories = self._get_report_categories().filtered(
+            lambda c: c.display_type in (display_type, "both")
+        )
+        if not include_other:
+            categories = categories.filtered(lambda c: not c.is_other)
+        return [(category.id, category.name) for category in categories]
```
AungKoKoLin1997: Can't we just return recordset instead of list of tuples?
If we must use tuples, you may just use `categories.name_get()`.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: After discussing this with Lin-san, we concluded that if we are not going to use post_init_hook, the current approach—executing the apply_report_category_defaults function during record creation and via an action—is more maintainable than using a default function, as it results in less code and better readability.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: I have deleted _get_report_categories().
Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: Now, I use the recordsets. Thank you

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: That's true. I have updated the code! Thank you!

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+    def _get_report_category_for_record(self, categories):
+        self.ensure_one()
+        other_category = categories.filtered("is_other")[:1]
+        matches = []
+        for category in categories.filtered(lambda c: not c.is_other):
+            if self.filtered_domain(category._get_domain()):
+                matches.append(category)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def apply_report_category_defaults(self):
```
AungKoKoLin1997: Would it be better?

```suggestion
    def assign_report_category(self):
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+        "svl.report.category",
+    )
+    report_category_changed = fields.Boolean(
+        compute="_compute_report_category_changed",
+        store=True,
+    )
+
+    @api.depends("report_category")
+    def _compute_report_category_changed(self):
+        categories = self.env["svl.report.category"].search([])
```
AungKoKoLin1997, line 25: Please add this inside `_get_report_category_for_record`.

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,50 @@
...
+    active = fields.Boolean(default=True)
+    sequence = fields.Integer(default=10)
+    domain = fields.Char()
+    is_other = fields.Boolean(string="Other Category")
+    display_type = fields.Selection(
+        [
+            ("storable", "Storable"),
+            ("consumable", "Consumable"),
+            ("both", "Both"),
+            ("invisible", "Invisible"),
```
AungKoKoLin1997: I don't get why we need this type and I don't see this is used in somewhere. Can you please explain me?

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def get_base_domain(self, wizard):
+        return [
+            ("actual_date", ">=", wizard.date_start),
+            ("actual_date", "<=", wizard.date_end),
+            ("product_id.active", "=", True),
+        ]
+
+    def get_product_categories(self):
+        category_objs = self.env["product.category"].search(
```
AungKoKoLin1997:
```suggestion
        categories = self.env["product.category"].search(
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,324 @@
...
+        self, display_type="storable", include_other=True
+    ):
+        categories = (
+            self.env["svl.report.category"]
+            .search([])
+            .filtered(lambda c: c.display_type in (display_type, "both"))
+        )
+        if not include_other:
+            categories = categories.filtered(lambda c: not c.is_other)
+        return categories
```
AungKoKoLin1997: Can you try like this?

```suggestion
    def get_inventory_operation_categories(
        self, display_type="storable", include_other=True
    ):
        domain = [("display_type", "in", [display_type, "both"])]
        if not include_other:
            domain.append(("is_other", "=", False))
        return self.env["svl.report.category"].search(domain)
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,324 @@
...
+            "storable": {
+                "header_left": _("Inventory Operation Type"),
+                "formula_column": "P",
+            },
+            "consumable": {
+                "header_left": _("Inventory Operation Type"),
+                "formula_column": "G",
+            },
+        }
+        return configs.get(report_type, configs["storable"])
```
AungKoKoLin1997: Is there any situation where no report_type is passed?

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
AungKoKoLin1997, line 1: It would be good if you reorder the view. (search, list, form, actions)

AungKoKoLin1997 (commented):
@nobuQuartile Could you take this chance to update the Copyright with our latest policy?

stock_valuation_layer_reporting/models/svl_report_category.py
nobuQuartile: The invisible is for the invisible report_category in the Excel.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: No. Should we remove configs["storable"]?

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
nobuQuartile: Done. Thank you

stock_valuation_layer_reporting/reports/svl_report.py
AungKoKoLin1997: If we don't have to consider the case, better to make the code clear. So, I prefer to remove.

stock_valuation_layer_reporting/models/svl_report_category.py
nobuQuartile: Quoted from Slack:
Toshikimi (Nobu) Shigenobu (QRTL)  [17:13]
display_type is selection. If we don't add the invisible value. We could not hide any stock valuation layer record in Excel.
[17:17]By the way, I will remove the display_type because I checked that all inventory valuation layer records from 2025 can be separated by report_category (Created based on the inventory_report’s domain).

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: Quoted from Slack:
Toshikimi (Nobu) Shigenobu (QRTL)  [18:06]
I think if we do it, we have to search for every records in assign_report_category().
Is that good?Aung Ko Ko Lin (QRTL)  [18:14]
You are right. So, ignore that suggestion but pls bring back compute method.

nobuQuartile:
Quoted from Slack:
Toshikimi (Nobu) Shigenobu (QRTL)  [12:14]
Do you know why the error occurs?
Aung Ko Ko Lin (QRTL)  [12:33]
Regarding the issue, I think the issue is from OCA tools. Even I tried to update the dotfiles, it still occurs. So, lets wait

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+        for record in self:
+            category_value = record._get_report_category_for_record(categories)
```
AungKoKoLin1997:
```suggestion
            category = record._get_report_category_for_record(categories)
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+        matches = []
+        for cat in categories.filtered(lambda c: not c.is_other):
+            if self.filtered_domain(cat._get_domain()):
+                matches.append(cat)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def assign_report_category(self):
+        categories = self.env["svl.report.category"].search([])
+        for record in self:
+            category_value = record._get_report_category_for_record(categories)
```
AungKoKoLin1997:
```suggestion
            category = record._get_report_category_for_record(categories)
```

AungKoKoLin1997 (approved):
Code Review: LGTM.
Just minor points.

stock_valuation_layer_reporting/wizards/svl_report_wizard_views.xml
```diff
@@ -0,0 +1,57 @@
...
+        <field name="view_id" ref="view_svl_report_wizard" />
+        <field name="target">new</field>
+    </record>
+    <menuitem
+        id="svl_report_menu"
+        name="SVL Report"
+        parent="stock.menu_warehouse_report"
+        action="svl_report_wizard_action"
+        sequence="30"
+        groups="account.group_account_manager,stock.group_stock_manager"
```
kanda999: This section is inconsistent with ir.model.access.
The account.group_account_manager menu item is visible, but the wizard displays an access error.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
kanda999, line 25:
```suggestion
        categories = self.env["svl.report.category"].search([('active', '=', True)])
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+        self.ensure_one()
+        other_category = categories.filtered("is_other")[:1]
+        matches = []
+        for cat in categories.filtered(lambda c: not c.is_other):
+            if self.filtered_domain(cat._get_domain()):
+                matches.append(cat)
+        return matches[0] if len(matches) == 1 else other_category
+
+    def assign_report_category(self):
+        categories = self.env["svl.report.category"].search([])
```
kanda999, line 66:
```suggestion
        categories = self.env["svl.report.category"].search([('active', '=', True)])
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,46 @@
...
+from odoo import api, fields, models
+
+
+class StockValuationLayer(models.Model):
+    _inherit = "stock.valuation.layer"
+
+    report_category = fields.Many2one(
+        "svl.report.category",
+    )
+    report_category_changed = fields.Boolean(
```
kanda999: The field name should be changed.
For example `report_category_manually_modified`

kanda999 (changes requested):

stock_valuation_layer_reporting/models/stock_valuation_layer.py
kanda999: Please ignore this

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,65 @@
...
+        <field name="name">stock.valuation.layer.tree.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.stock_valuation_layer_tree" />
+        <field name="arch" type="xml">
+            <field name="product_id" position="after">
+                <field name="report_category" groups="base.group_system" readonly="0" />
+            </field>
+        </field>
+    </record>
+        <record
```
kanda999:
```suggestion
    <record
```
Please align the indentation.

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,69 @@
...
+        id="stock_valuation_layer_search_inherit_stock_valuation_layer_reporting"
+        model="ir.ui.view"
+    >
+        <field name="name">stock.valuation.layer.search.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.view_inventory_valuation_search" />
+        <field name="arch" type="xml">
+            <filter name="has_remaining_qty" position="after">
+                <filter
+                    string="Report Category Changed"
```
kanda999:
```suggestion
                    string="Report Category Manual Modified"
```

stock_valuation_layer_reporting/readme/CONFIGURE.md
```diff
@@ -0,0 +1,14 @@
+Configuration
+
+1. Go to Inventory > Configuration > SVL Report Categories.
+2. Create categories with:
+   - Name
+   - Sequence
+   - Display Type (storable/both/invisible)
```
AungKoKoLin1997: @nobuQuartile I think you need to update that with the latest design.

stock_valuation_layer_reporting/readme/CONFIGURE.md
```diff
@@ -4,11 +4,16 @@ Configuration
...
    - Domain (stock.valuation.layer domain)
-   - Other Category (set is_other for the fallback category)
+   - Other Category (set `is_other` for the fallback category)
 
 Other category behavior
 
-- If multiple categories match a record, the Other category is assigned.
-- If no categories match a record, the Other category is assigned.
+1. Please assign Other category to only one category.
+2. If exactly one category matches a record, that category is assigned.
```
kanda999: I think it's strange that this sentence is here. I think it should be merged with sentence 3.

nobuQuartile: Done

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                )
+                ws.write(row, 7, valuation.product_id.name)
+                ws.write(row, 8, valuation.product_id.type)
+                ws.write(row, 9, valuation.product_id.categ_id.name)
+                ws.write(row, 10, valuation.stock_move_id.location_id.name)
+                ws.write(row, 11, valuation.stock_move_id.location_dest_id.name)
+                ws.write(row, 12, valuation.quantity)
+                ws.write(row, 13, valuation.uom_id.name)
+                ws.write(row, 14, valuation.product_id.categ_id.property_cost_method)
+                ws.write(row, 15, valuation.value),
```
kanda999:
```suggestion
                ws.write(row, 15, valuation.value)
```
Is there any meaning to this comma?

kanda999 (approved):
Functioanl check and code review

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def generate_valuation_report(self, workbook, wizard):
+        categories = self.env["product.category"].search(
+            [("is_report_category", "=", True)]
+        )
+        self.setup_summary_sheet(workbook, categories.mapped("name"), "valuation")
+        for category in categories:
+            ws = workbook.add_worksheet(category.name)
+
+            # Write the header
```
yostashiro: Can we remove blank lines and comments that don't seem to carry additional context?
```suggestion
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+        self, display_type="storable", include_other=True
+    ):
+        domain = [("display_type", "in", [display_type, "both"])]
+        if not include_other:
+            domain.append(("is_other", "=", False))
+        return self.env["svl.report.category"].search(domain)
+
+    def get_valuation_domain(self, category, wizard):
+        return [
+            ("product_id.active", "=", True),
```
yostashiro: Shouldn't we remove this?

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                _("Last Purchase Accounting Date"),
+            ]
+            column_widths = [25, 20, 10, 20, 12, 15, 30]
+            for col, width in enumerate(column_widths):
+                ws.set_column(col, col, width)
+            for col, header in enumerate(headers):
+                ws.write(0, col, header)
+
+            # Fetch the valuation layers for the product category and date range
+            valuation_obj = self.env["stock.valuation.layer"]
```
yostashiro: Avoid defining a variable that's only used once.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+        if html_content:
+            soup = BeautifulSoup(html_content, "html.parser")
+            return soup.get_text()
+        return False
+
+    def get_base_domain(self, wizard):
+        return [
+            ("actual_date", ">=", wizard.date_start),
+            ("actual_date", "<=", wizard.date_end),
+            ("product_id.active", "=", True),
```
yostashiro: Why should we have this condition?

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                    valuation.stock_move_id.analytic_account_names or "",
+                )
+
+    def generate_summary_report(self, workbook, wizard):
+        ws = workbook.add_worksheet(_("SVL Report Summary"))
+        valuation_obj = self.env["stock.valuation.layer"]
+        product_categories = self.env["product.category"].search(
+            [("is_report_category", "=", True)]
+        )
+        base_storable_domain = expression.AND(
```
yostashiro, line 245:
```suggestion
        storable_domain = expression.AND(
```

stock_valuation_layer_reporting/readme/CONFIGURE.md
```diff
@@ -0,0 +1,18 @@
+Configuration
+
```
yostashiro:
```suggestion
```

stock_valuation_layer_reporting/models/svl_report_category.py
yostashiro, line 1: Can we ship the module with category records (noupdate=True) to avoid manual operation?

stock_valuation_layer_reporting/readme/CONFIGURE.md
```diff
@@ -0,0 +1,18 @@
...
+2. Create categories with:
+   - Name
+   - Sequence
+   - Display Type (Storable/Consumable/Both)
+   - Domain (stock.valuation.layer domain)
+   - Other Category (set `is_other` for the fallback category)
+
+Other Category behavior
+
+1. Please assign Other Category to only one category.
```
yostashiro: Add a constraint?

stock_valuation_layer_reporting/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Quartile Limited (https://www.quartile.co)
```
yostashiro:
```suggestion
# Copyright 2026 Quartile (https://www.quartile.co)
```

stock_valuation_layer_reporting/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Stock Valuation Layer Reporting",
+    "version": "16.0.1.0.0",
+    "author": "Quartile Limited",
```
yostashiro:
```suggestion
    "author": "Quartile",
```

stock_valuation_layer_reporting/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_svl_report_wizard_stock_manager,svl.report.wizard,model_svl_report_wizard,stock.group_stock_manager,1,1,1,1
```
yostashiro, line 5: This is too loose, I think. It should probably be `base.group_system`.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                domain, fields_to_aggregate, ["product_id"]
+            )
+
+            # Write the aggregated data
+            row = 1
+            for valuation_data in valuation_grouped_data:
+                product = self.env["product.product"].browse(
+                    valuation_data["product_id"][0]
+                )
+                company_currency = self.env.company.currency_id
```
yostashiro: Keep this outside the loop.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                unit_cost = float_round(
+                    valuation_data["value"] / valuation_data["quantity"]
+                    if valuation_data["quantity"] > 0
+                    else 0,
+                    precision_rounding=company_currency.rounding,
+                    rounding_method="UP",
+                )
```
yostashiro, line 126: To handle the cases where quantity is negative. It's not supposed to happen but it's better to be able to capture it in case it happens.
```suggestion
                qty = valuation_data["quantity"]
                unit_cost = float_round(
                    valuation_data["value"] / qty
                    if not float_is_zero(qty, precision_rounding=product.uom_id.rounding)
                    else 0,
                    precision_rounding=company_currency.rounding,
                    rounding_method="UP",
                )
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                ws.write(row, 0, product.name)
+                ws.write(row, 1, product.default_code)
```
yostashiro:
```suggestion
                ws.write(row, 1, product.default_code or "")
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+
+            # Write the data to the worksheet
+            for row, valuation in enumerate(valuations, start=1):
+                actual_date = (
+                    valuation.actual_date
+                    if report_type == "storable"
+                    else valuation.stock_move_id.actual_date
+                )
+                ws.write(row, 0, valuation.reference)
+                ws.write(row, 1, valuation.stock_move_id.origin)
```
yostashiro:
```suggestion
                ws.write(row, 1, valuation.stock_move_id.origin or "")
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                ws.write(row, 2, actual_date.strftime("%Y-%m-%d"))
+                ws.write(
+                    row,
+                    3,
+                    self.parse_html(valuation.stock_move_id.picking_id.note) or "",
+                )
+                ws.write(row, 4, valuation.create_uid.name)
+                ws.write(row, 5, valuation.stock_move_id.picking_id.partner_id.name)
```
yostashiro:
```suggestion
                ws.write(row, 5, valuation.stock_move_id.picking_id.partner_id.name or "")
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
+from bs4 import BeautifulSoup
```
yostashiro: Add the license header.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+
+    def _get_report_category_for_record(self, categories):
+        self.ensure_one()
+        other_category = categories.filtered("is_other")[:1]
+        matches = []
+        for cat in categories.filtered(lambda c: not c.is_other):
+            if self.filtered_domain(cat._get_domain()):
+                matches.append(cat)
+        return matches[0] if len(matches) == 1 else other_category
```
yostashiro: This looks a bit risky. Prioritize the first matching one, or log a warning?

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,69 @@
...
+        <field name="name">stock.valuation.layer.form.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.stock_valuation_layer_form" />
+        <field name="arch" type="xml">
+            <field name="product_id" position="after">
+                <field
+                    name="report_category_id"
+                    groups="base.group_system"
+                    readonly="0"
+                    force_save="1"
```
yostashiro: Do we need this force_save?

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,69 @@
...
+        <field name="name">stock.valuation.layer.tree.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.stock_valuation_layer_tree" />
+        <field name="arch" type="xml">
+            <field name="product_id" position="after">
+                <field
+                    name="report_category_id"
+                    groups="base.group_system"
+                    readonly="0"
+                />
```
yostashiro, line 34: Just an idea for usability improvement (add a boolean field report_category_manually_modified to the model).
```suggestion
                <field name="report_category_manually_modified" invisible="1" />
                <field
                    name="report_category_id"
                    groups="base.group_system"
                    readonly="0"
                    decoration-info="report_category_manually_modified"
                />
```

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,69 @@
...
+    >
+        <field name="name">stock.valuation.layer.tree.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.stock_valuation_layer_tree" />
+        <field name="arch" type="xml">
+            <field name="product_id" position="after">
+                <field
+                    name="report_category_id"
+                    groups="base.group_system"
+                    readonly="0"
```
yostashiro: What is `readonly="0"` for?
```suggestion
                    groups="stock.group_stock_manager"
                    optional="hide"
```

stock_valuation_layer_reporting/views/stock_valuation_layer_views.xml
```diff
@@ -0,0 +1,69 @@
...
+    >
+        <field name="name">stock.valuation.layer.form.svl.reporting</field>
+        <field name="model">stock.valuation.layer</field>
+        <field name="inherit_id" ref="stock_account.stock_valuation_layer_form" />
+        <field name="arch" type="xml">
+            <field name="product_id" position="after">
+                <field
+                    name="report_category_id"
+                    groups="base.group_system"
+                    readonly="0"
```
yostashiro:
```suggestion
```

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,49 @@
...
+    _name = "svl.report.category"
+    _description = "SVL Report Category"
+    _order = "sequence, id"
+
+    name = fields.Char(required=True)
+    active = fields.Boolean(default=True)
+    sequence = fields.Integer(default=10)
+    domain = fields.Char()
+    is_other = fields.Boolean(string="Other Category")
+    display_type = fields.Selection(
```
yostashiro:
```suggestion
    report_type = fields.Selection(
```
This may be a better fit? `display_type` in Odoo's terminology usually means something else and could be misleading.

stock_valuation_layer_reporting/models/product_category.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Quartile Limited (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductCategory(models.Model):
+    _inherit = "product.category"
+
+    is_report_category = fields.Boolean(string="Include in SVL Report")
```
yostashiro:
```suggestion
    is_svl_report_category = fields.Boolean(string="Include in SVL Report")
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,319 @@
...
+                ws.write(row, 4, unit_cost)
+                ws.write(row, 5, valuation_data["value"])
+
+                # Convert the date to the desired format (YYYY-MM-DD)
+                last_purchase_date = fields.Date.from_string(product.last_purchase_date)
+                if last_purchase_date:
+                    ws.write(row, 6, last_purchase_date.strftime("%Y-%m-%d"))
+                row += 1
+
+    def setup_storable_worksheet_headers(self, ws):
```
yostashiro:
```suggestion
    def setup_stock_operation_worksheet_headers(self, ws):
```

yostashiro (changes requested):
Please confirm the validity of my comments.

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: We should not change the name.
In the for-loop, base_storable_domain is reassigned to storable_domain, so after renaming base_storable_domain to storable_domain, the second and subsequent categories inherit the previous category’s domain.

nobuQuartile:
@yostashiro 
Please continue your review.

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,69 @@
...
+                continue
+            existing = self.search(
+                [
+                    ("id", "!=", record.id),
+                    ("is_other", "=", True),
+                ],
+                limit=1,
+            )
+            if existing:
+                raise UserError(
```
yostashiro:
```suggestion
                raise ValidationError(
```

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,69 @@
...
+    def _check_unique_other_category(self):
+        for record in self:
+            if not record.is_other or not record.active:
+                continue
+            existing = self.search(
+                [
+                    ("id", "!=", record.id),
+                    ("is_other", "=", True),
+                ],
+                limit=1,
```
yostashiro, line 61:
```suggestion
                [("id", "!=", record.id), ("is_other", "=", True)], limit=1,
```

nobuQuartile:
Done functional test in Copy.

Tested areas:
・Recalculation of report categories on existing SVLs and manual apply action
・SVL report printing
・Stock receipt without error
・Prevention of creating two Other Category
・Inventory managers can view but not update the report category on SVLs

nobuQuartile:
The user decided that when multiple categories match, the Other category should be assigned. To address the previously raised concern about losing visibility into what caused the multiple matches, an error was logged for later review.

stock_valuation_layer_reporting/models/svl_report_category.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+            if existing:
+                raise ValidationError(
+                    _(
+                        "Only one 'Other Category' is allowed per. "
```
yostashiro: This message is incomplete.

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,306 @@
...
+            self.generate_inventory_operation_report(
+                workbook, wizard, data["report_type"]
+            )
+        else:
+            self.generate_summary_report(workbook, wizard)
+
+    def parse_html(self, html_content):
+        if html_content:
+            soup = BeautifulSoup(html_content, "html.parser")
+            return soup.get_text()
```
yostashiro: We'd technically need to define an external dependency in the manifest, otherwise.
```suggestion
            return html2plaintext(html_content)
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,306 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from bs4 import BeautifulSoup
+
+from odoo import _, fields, models
+from odoo.osv import expression
+from odoo.tools import float_round
```
yostashiro:
```suggestion
from odoo import _, fields, models
from odoo.osv import expression
from odoo.tools import float_round, html2plaintext
```

stock_valuation_layer_reporting/readme/CONFIGURE.md
```diff
@@ -0,0 +1,27 @@
...
+   - Report Type (Storable/Consumable/Both)
+   - Domain (stock.valuation.layer domain)
+   - Other Category (set `is_other` for the fallback category)
+
+## Other Category Behavior
+
+1. Please assign Other Category to only one category (enforced by system).
+2. If exactly one category matches, assign that category.
+3. If multiple categories match, an error is logged and the first matching
+   category (by sequence) is assigned.
```
yostashiro: This contradicts with the next comment.

stock_valuation_layer_reporting/views/svl_report_category_views.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="svl_report_category_form" model="ir.ui.view">
```
yostashiro:
```suggestion
    <record id="svl_report_category_view_form" model="ir.ui.view">
```

stock_valuation_layer_reporting/views/svl_report_category_views.xml
```diff
@@ -0,0 +1,52 @@
...
+                            widget="domain"
+                            options="{'model': 'stock.valuation.layer'}"
+                            attrs="{'invisible': [('is_other', '=', True)]}"
+                        />
+                    </group>
+                </sheet>
+            </form>
+        </field>
+    </record>
+    <record id="svl_report_category_tree" model="ir.ui.view">
```
yostashiro:
```suggestion
    <record id="svl_report_category_view_tree" model="ir.ui.view">
```

stock_valuation_layer_reporting/wizards/svl_report_category_wizard_views.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="svl_report_category_wizard_form" model="ir.ui.view">
```
yostashiro:
```suggestion
    <record id="svl_report_category_wizard_view_form" model="ir.ui.view">
```

stock_valuation_layer_reporting/wizards/svl_report_wizard_views.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_svl_report_wizard" model="ir.ui.view">
```
yostashiro, line 3:
```suggestion
    <record id="svl_report_wizard_view_form" model="ir.ui.view">
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,70 @@
...
+        if len(matches) == 1:
+            return matches[0]
+        if len(matches) > 1:
+            _logger.error(
+                "SVL(%s: %s) matched multiple report categories %s",
+                self.reference,
+                self.product_id.display_name or "",
+                matches.mapped("name"),
+            )
+        return other_category
```
yostashiro, line 63: No logger for the multiple matching cases?

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,306 @@
...
+            fields_to_aggregate = ["quantity", "value"]
+            valuation_grouped_data = self.env["stock.valuation.layer"].read_group(
+                domain, fields_to_aggregate, ["product_id"]
+            )
+            company_currency = self.env.company.currency_id
+            row = 1
+            for valuation_data in valuation_grouped_data:
+                product = self.env["product.product"].browse(
+                    valuation_data["product_id"][0]
+                )
```
yostashiro: For better performance.
```suggestion
            company_currency = self.env.company.currency_id
            product_ids = [d["product_id"][0] for d in valuation_grouped_data]
            product_map = {
                p.id: p for p in self.env["product.product"].browse(product_ids)
            }
            row = 1
            for valuation_data in valuation_grouped_data:
                product = product_map[valuation_data["product_id"][0]]
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
nobuQuartile: There is the logger.
https://github.com/qrtl/axls-custom/blob/cbd7d9b4c1d80276f5ab44a782650cf1df2e89d7/stock_valuation_layer_reporting/models/stock_valuation_layer.py#L49-L55

nobuQuartile:
Done functional review.
I have checked the no-match log and the multiple-match log in Staging.

nobuQuartile:
@yostashiro 
The code has passed Claude's review.

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,83 @@
...
+        matches = []
+        for cat in categories.filtered(lambda c: not c.is_other):
+            try:
+                if self.filtered_domain(cat._get_domain()):
+                    matches.append(cat)
+            except Exception as e:
+                _logger.error(
+                    "SVL(%s: %s) failed to evaluate domain for report category %s: %s",
+                    self.reference,
+                    self.product_id.display_name or "",
```
yostashiro: `or ""` is unnecessary.
```suggestion
                    self.product_id.display_name,
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,83 @@
...
+                    cat.name,
+                    e,
+                )
+        if len(matches) == 1:
+            return matches[0]
+        if len(matches) == 0:
+            _logger.warning(
+                "SVL(%s: %s) matched no report categories",
+                self.reference,
+                self.product_id.display_name or "",
```
yostashiro:
```suggestion
                self.product_id.display_name,
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+        else:
+            _logger.warning(
+                "SVL(%s: %s) matched multiple report categories %s",
+                self.reference,
+                self.product_id.display_name or "",
```
yostashiro:
```suggestion
                self.product_id.display_name,
```

stock_valuation_layer_reporting/models/stock_valuation_layer.py
```diff
@@ -0,0 +1,83 @@
...
+        category_records = {}
+        for record in self:
+            category = record._get_report_category_for_record(categories)
+            if not category:
+                continue
+            cat_id = category.id
+            if cat_id not in category_records:
+                category_records[cat_id] = record
+            else:
+                category_records[cat_id] |= record
```
yostashiro: Can be made more compact with defaultdict.
```suggestion
        category_records = defaultdict(lambda: self.env["stock.valuation.layer"])
        for record in self:
            category = record._get_report_category_for_record(categories)
            if not category:
                continue
            category_records[category.id] |= record
```

stock_valuation_layer_reporting/reports/svl_report.py
```diff
@@ -0,0 +1,300 @@
...
+            product_ids = [d["product_id"][0] for d in valuation_grouped_data]
+            product_map = {
+                p.id: p for p in self.env["product.product"].browse(product_ids)
+            }
+            row = 1
+            for valuation_data in valuation_grouped_data:
+                product = product_map[valuation_data["product_id"][0]]
+                qty = valuation_data["quantity"]
+                unit_cost = float_round(
+                    valuation_data["value"] / qty if qty else 0,
```
yostashiro: Why did you remove `float_is_zero()` check on qty? I thought it'd be good to have it.

yostashiro (approved):

stock_valuation_layer_reporting/reports/svl_report.py
nobuQuartile: to reduce "from odoo.tools.float_utils import float_is_zero"
I think "if qty" works well too.

nobuQuartile:
I have done functional test again.

PR #142, opened 2026-03-10
[6424][IMP] account_move_obc_csv

kanda999:
[6424](https://www.quartile.co/web#id=6424&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)

### DIFF
--- a/account_move_obc_csv/report/report_account_move_obc_csv.py
+++ b/account_move_obc_csv/report/report_account_move_obc_csv.py
@@ -89,8 +89,10 @@ class AccountMoveObcCsv(models.AbstractModel):
         """Get the account code to use for OBC export.
         For vendor bill lines using the purchase accrual (Goods Received Not Invoiced)
         account, return the stock valuation account instead to represent it as
-        "material purchase" in the OBC system."""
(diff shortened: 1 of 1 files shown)


yostashiro (approved):
I'm not in favor of increasing the method args, though. I think there is probably a cleaner way.

AungKoKoLin1997 (commented):
In my opinion, the fix can be done by adjusting two or three lines.
Inside `_get_line_analytic_account` use `if line.account_id.code != account_code` instead of `if is_substituted`. Then, just pass `account_code` parameter from `_update_vals` when you call `_get_line_analytic_account`. Or, you can get rid of adding extra `_get_line_analytic_account` and add above condition inside `_update_vals` like below.


```
if line.account_id.code != account_code:
    project = self.env["account.analytic.account"]
    department = self.env["account.analytic.account"]
else:
    project = line.analytic_line_ids.filtered(
            lambda x: x.plan_type == "project"
        ).account_id[:1]
    department = line.analytic_line_ids.filtered(
        lambda x: x.plan_type == "department"
    ).account_id[:1]
```

PR #147, opened 2026-05-16
[6131][ADD] purchase_deposit_currency

kanda999:
[6131](https://www.quartile.co/web#id=6131&menu_id=505&cids=3&action=1457&model=project.task&view_type=form)

### DIFF
--- a/purchase_deposit_currency/__manifest__.py
+++ b/purchase_deposit_currency/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "Purchase Deposit Multi-Currency",
-    "version": "16.0.2.1.0",
+    "version": "16.0.2.2.0",
     "author": "Quartile Limited",
     "website": "https://www.quartile.co",
     "category": "Purchase",
--- a/purchase_deposit_currency/__manifest__.py
+++ b/purchase_deposit_currency/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 {
     "name": "Purchase Deposit Multi-Currency",
-    "version": "16.0.2.0.0",
+    "version": "16.0.2.1.0",
     "author": "Quartile Limited",
     "website": "https://www.quartile.co",
     "category": "Purchase",
--- a/purchase_deposit_currency/__manifest__.py
+++ b/purchase_deposit_currency/__manifest__.py
@@ -1,0 +1,20 @@
+# Copyright 2026 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Purchase Deposit Multi-Currency",
+    "version": "16.0.2.0.0",
+    "author": "Quartile Limited",
+    "website": "https://www.quartile.co",
+    "category": "Purchase",
+    "license": "AGPL-3",
+    "summary": "Let each vendor-bill line carry a manually-entered "
+    "company-currency amount so foreign-currency deposits and invoices "
+    "post the exact JPY (company-currency) value the user actually paid, "
+    "bypassing Odoo's exchange-rate conversion. The deposit value is "
+    "carried over to the deposit-offset line of the final invoice.",
+    "depends": ["purchase_deposit"],
+    "data": [
+        "views/account_move_views.xml",
+    ],
+    "installable": True,
+}
--- a/purchase_deposit_currency/models/account_move.py
+++ b/purchase_deposit_currency/models/account_move.py
@@ -1,10 +1,96 @@
 # Copyright 2026 Quartile Limited
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
-from odoo import models
+from odoo import api, models
 
 
 class AccountMove(models.Model):
     _inherit = "account.move"
+
+    def _moves_needing_deposit_company_adjustment(self):
+        return self.filtered(
+            lambda m: m.move_type in ("in_invoice", "in_refund")
+            and m.currency_id != m.company_id.currency_id
+            and m.line_ids.filtered(
+                lambda l: l.display_type == "product"
+                and l.purchase_line_id.is_deposit
+                and l.purchase_line_id.deposit_company_amount
+                and l.quantity < 0
+            )
+        )
+
+    def _apply_deposit_company_adjustment(self):
+        """Push the deposit's exchange-rate difference onto the product
+        line(s) of the final invoice.
+
+        The deposit-offset line's company-currency balance is pinned to the
+        JPY actually paid (``deposit_company_amount``), while the product
+        lines are booked at the current rate. The gap between the deposit
+        valued at the current rate and the JPY actually paid is the
+        rate-difference; it belongs in the goods' acquisition cost. We absorb
+        it into the product lines' ``company_amount`` so the goods value (and
+        SVL) reflects the true cost and the auto-balanced payable equals the
+        remaining foreign amount at the current rate.
+        """
+        for move in self:
+            company_currency = move.company_id.currency_id
+            offset_lines = move.line_ids.filtered(
+                lambda l: l.display_type == "product"
+                and l.purchase_line_id.is_deposit
+                and l.purchase_line_id.deposit_company_amount
+                and l.quantity < 0
+            )
+            # delta = deposit-at-current-rate - JPY actually paid (pinned
+            # balance). Signed, so it works for in_invoice and in_refund.
+            total_delta = sum(
+                offset._deposit_natural_balance() - offset.balance
+                for offset in offset_lines
+            )
+            if company_currency.is_zero(total_delta):
+                continue
+            product_lines = move.line_ids.filtered(
+                lambda l: l.display_type == "product"
+                and l.purchase_line_id
+                and not l.purchase_line_id.is_deposit
+                # Auto-fill empty lines and lines we filled before (rate may
+                # have changed); never clobber a manual override.
+                and (not l.company_amount or l.deposit_amount_adjusted)
+            )
+            if not product_lines:
+                continue
+            naturals = {l.id: l._deposit_natural_balance() for l in product_lines}
+            total_weight = sum(abs(v) for v in naturals.values())
+            line_count = len(product_lines)
+            remaining = total_delta
+            for idx, line in enumerate(product_lines):
+                if idx < line_count - 1:
+                    if total_weight:
+                        weight = abs(naturals[line.id]) / total_weight
+                        raw_share = total_delta * weight
+                    else:
+                        raw_share = total_delta / line_count
+                    share = company_currency.round(raw_share)
+                    remaining -= share
+                else:
+                    # last line absorbs the rounding remainder
+                    share = company_currency.round(remaining)
+                target = company_currency.round(naturals[line.id] + share)
+                line.with_context(skip_deposit_company_adjustment=True).write(
+                    {"company_amount": target, "deposit_amount_adjusted": True}
+                )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        moves = super().create(vals_list)
+        target = moves._moves_needing_deposit_company_adjustment()
+        target._apply_deposit_company_adjustment()
+        return moves
+
+    def write(self, vals):
+        res = super().write(vals)
+        if not self.env.context.get("skip_deposit_company_adjustment"):
+            target = self._moves_needing_deposit_company_adjustment()
+            target._apply_deposit_company_adjustment()
+        return res
 
     def action_post(self):
         """When a deposit vendor bill is posted, capture the company-currency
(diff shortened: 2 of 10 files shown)


AungKoKoLin1997:
@kanda999 Please fix the pre-commit issue.

purchase_deposit_currency/__manifest__.py
AungKoKoLin1997, line 1: Please reformat as per our standard and I think summary should be one line brief description.

purchase_deposit_currency/models/account_move.py
```diff
@@ -0,0 +1,28 @@
...
+        """
+        deposit_writes = []
+        for move in self:
+            for line in move.line_ids:
+                po_line = line.purchase_line_id
+                if not po_line or not po_line.is_deposit:
+                    continue
+                amount = line.company_amount or abs(line.balance)
+                if amount:
+                    deposit_writes.append((po_line, amount))
```
AungKoKoLin1997: Doesn't it consider both deposit bill and normal vendor bill?

AungKoKoLin1997:
@kanda999 @nobuQuartile This PR is ready to review.

purchase_deposit_company_amount/models/account_move_line.py
```diff
@@ -0,0 +1,117 @@
...
+                    "'%(line)s' of '%(move)s' the standard exchange-rate "
+                    "conversion applies; clear the value to continue."
+                )
+                % {
+                    "field": line._fields["company_amount"].string,
+                    "line": line.name or line.product_id.display_name or "/",
+                    "move": line.move_id.display_name,
+                }
+            )
+
```
nobuQuartile, line 52: Shouldn't we also prevent users from entering this field when the currency is the company currency?

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,168 @@
...
+                )
+            )
+
+    def _get_deposit_offset_lines(self):
+        """The negative-quantity deposit lines ``purchase_deposit`` adds to the
+        final vendor bill to net off a deposit already billed.
+        """
+        self.ensure_one()
+        return self.line_ids.filtered(
+            lambda l: l.display_type == "product"
```
nobuQuartile, line 67: Do we need this line?

purchase_deposit_company_amount/models/account_move_line.py
nobuQuartile: I know we do not have to do this. but it is more user friendly.

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,168 @@
...
+        therefore belongs in the acquisition cost, not in an FX gain or loss.
+        Product lines the user has priced by hand are left out of it -- a
+        manual value is never overwritten.
+        """
+        self.ensure_one()
+        company_currency = self.company_id.currency_id
+        if self.currency_id == company_currency:
+            return {}
+        product_lines = self.line_ids.filtered(lambda l: l.display_type == "product")
+        if not product_lines.filtered(lambda l: l.purchase_line_id.is_deposit):
```
nobuQuartile, line 132: Why don't you use move's is_deposit

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+
+    def _get_company_amount_targets(self):
+        """Company-currency balance every overridden line of this move should
+        end up with, as ``{line: balance}``. Lines absent from the result keep
+        the standard rate-based conversion.
+
+        Three sources, in precedence order:
+
+        1. a manual ``company_amount`` -- the amount the user says was paid;
```
nobuQuartile, line 114:
```suggestion
1. a manual ``company_amount`` on the deposit line of a deposit bill --
   the company-currency amount the user says was actually paid;

```

nobuQuartile (commented):
Partial review from me

nobuQuartile:
@AungKoKoLin1997 
This review is from odoo-review.
I confirmed this issue occurs in the Test environment.
By the way, you are doing well.
We're almost there, let's keep going!

**Critical**

Rewrites balance on posted journal entries — models/account_move.py:157-168

_apply_company_amount_overrides() has no state guard. account.move.write() enters _sync_dynamic_lines() → line_ids._sync_invoice() regardless of state (addons/account/models/account_move.py:2360, same:2213-2215), so any write against a posted bill — editing the reference or due date, button_draft, a write from another module — recomputes the targets and runs line.balance = target.

There are two realistic ways the targets end up differing from what was posted:

A currency rate added or corrected for the bill's date. _get_rate_based_balance() depends on currency_rate, which comes from the rate table, so a rate landing on that date after the bill was posted changes delta and leaves the goods lines' targets out of step with the posted balances. This repo has currency_rate_update_mizuho feeding rates in by cron, so posting a bill dated today before the cron runs is enough to trigger it.

A deposit bill reset to draft, corrected and re-posted — specifically when deposit_company_amount ends up at a different non-zero value (resetting it to zero is safe, since the targets then come out empty).

Either outcome is bad:

If the entry is reconciled, balance is a reconciliation-protected field (addons/account/models/account_move_line.py:2815 → same:1261-1266), so a UserError ("You cannot do this modification on a reconciled journal entry") blocks an otherwise unrelated operation. Inside a lock period, _check_fiscalyear_lock_date() / _check_tax_lock_date() fail the same way.
If it is not reconciled, the posted entry's amounts are silently rewritten while the stock.valuation.layer records and price-difference journal items created at posting time keep the old values.
Suggested fix: in _sync_invoice, capture each affected move's state before the yield (i.e. before the write lands) and skip the moves that were already posted. Guarding simply on move.state != "draft" would also skip during write({'state': 'posted'}), but that should be harmless for vendor bills: _post requires invoice_date (addons/account/models/account_move.py:3564-3568), so the rate cannot move at posting time and the balances pinned in draft carry through unchanged. Worth confirming with a test.

**Major**

Clearing company_amount does not restore the standard conversion — models/account_move.py:45-94, models/account_move_line.py:13-24

Both the field's help text and readme/USAGE.md state that leaving the field empty keeps the standard conversion, but it does not come back. Clearing the value empties product_lines.filtered("company_amount"), so targets is {} and _apply_company_amount_overrides() bails out at if not targets: continue. Core's _sync_invoice will not re-derive the balance either, because none of amount_currency, currency_rate or move_type changed (addons/account/models/account_move_line.py:1436-1443). The result is that the balance pinned at 3900 — and the payable line as adjusted by _rebalance_payment_term_lines() — stay in place, and the user can post that.

Suggested fix: give every eligible line an explicit target of _get_rate_based_balance() when its company_amount is empty, i.e. express "no override" as "back to the rate-converted value" rather than as absence from targets. A single test stating the documented behaviour would have caught this.

AungKoKoLin1997:
@nobuQuartile I understand the scenario you showed me.
But what I am wondering is this can happen in real operation.
After deposit bill and vendor bill are posted, user change the deposit bill company amount by doing reset to draft.
I understand they can make mistake in deposit bill and update but that can only happen before creating final vendor bill or before posting final vendor bill.

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -31,6 +36,28 @@ def _compute_is_deposit(self):
...
+        ``_get_company_amount_targets`` skips the move entirely. Accepting a
+        value there would silently do nothing, so the field is refused and its
+        column hidden.
+        """
+        for move in self:
+            move.allow_company_amount = bool(
+                move.is_deposit and move.currency_id != move.company_currency_id
+            )
+
+    @api.constrains("currency_id", "line_ids")
```
nobuQuartile: @api.constrains("is_deposit", "line_ids")
Why don't you use is_deposit?

purchase_deposit_company_amount/models/account_move_line.py
```diff
@@ -19,26 +19,29 @@ class AccountMoveLine(models.Model):
...
-        purchase order line with a negative one) and so is every other bill.
+        ``allow_company_amount`` is enough on its own: it looks for a deposit
+        line with a positive quantity, so the final bill is out (its offset
+        reuses the same purchase order line with a negative one) and so is
+        every other bill, and it requires a foreign currency, without which
+        there is no conversion to override in the first place.
         """
         for line in self.filtered("company_amount"):
-            if line.move_id.is_deposit:
+            if line.move_id.allow_company_amount:
```
nobuQuartile, line 37: @api.constrains("company_amount", "move_id", "quantity")
def _check_company_amount_allowed(self):
    for line in self.move_id.line_ids.filtered("company_amount"):
        if line.move_id.allow_company_amount:
            continue
        raise ValidationError(...)
----------------------------------------------------------

### `_check_company_amount_allowed` doesn't catch the quantity case

`models/account_move.py:52-59` — the docstring says the move constraint catches
"the deposit line losing its positive quantity". It doesn't.

`_validate_fields` (odoo/models.py:1364) only runs constraints matching the fields
written, on the record written. Writing `quantity` on `account.move.line` fires
neither constraint: the line one lists `("company_amount", "move_id")`, and
`account.move.line.write` never calls `move_id._validate_fields(['line_ids'])`.

`_get_company_amount_targets:135` selects on `company_amount` alone and never
consults `allow_company_amount`, so the override keeps applying after the gate
has closed. With `quantity = 0` the line ends up **balance 3900 / amount_currency 0**
— it passes `check_amount_currency_balance_sign`, the payment term rebalances,
the move posts, and the column is now hidden so nobody can see or clear the 3900.

AungKoKoLin1997: @nobuQuartile I think these points are not necessary to fix and we don't need to consider that.
Have you checked yourself first? Please let me know if you have different opinion.

nobuQuartile: @AungKoKoLin1997 
Sorry, it does not occur.

nobuQuartile:
### Not in scope: `company_amount` accepted on non-deposit lines

`_check_company_amount_allowed` and `_get_company_amount_targets` gate on the
move (`allow_company_amount`), not on the line, so on a deposit bill the
override would also be honoured on a product line other than the deposit
product.

In practice this does not occur: every deposit bill raised by the Register
Deposit wizard carries the deposit product line only — we have no case of a
deposit bill containing any other product. Reaching the state requires manually
adding a line to the deposit bill, which is not part of the flow this module
supports. I have checked that the client does not add other lines for deposit invoices in Copy.

Treating it as out of scope for this PR.

nobuQuartile (approved):
LGTM: Functional review and code review

purchase_deposit_company_amount/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2026 Quartile (https://www.quartile.co)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Purchase Deposit Company Amount",
+    "summary": "Book purchase deposits at the company-currency amount actually paid",
+    "version": "16.0.1.0.0",
+    "author": "Quartile Limited",
```
yostashiro:
```suggestion
    "author": "Quartile",
```

purchase_deposit_company_amount/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://www.quartile.co",
+    "category": "Purchase",
+    "license": "AGPL-3",
+    "maintainers": ["kanda999"],
+    # purchase_stock: holds the price-difference logic that consumes
+    # ``_get_gross_unit_price`` (which is itself defined in stock_account) and
+    # writes the stock valuation adjustment the override has to reach.
+    "depends": ["purchase_deposit", "purchase_stock"],
```
yostashiro:
```suggestion
    "depends": ["purchase_stock", "purchase_deposit"],
```

purchase_deposit_company_amount/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,15 @@
+This module does the following:
```
yostashiro:
```suggestion
This module books a foreign-currency purchase deposit at the company-currency amount
actually paid and carries that figure through to the final bill, so the deposit account
clears to zero and the exchange-rate difference lands in the inventory cost of the goods
instead of being stranded on the balance sheet.

In detail:
```

purchase_deposit_company_amount/models/account_move.py
```diff
@@ -0,0 +1,239 @@
...
+        ``currency_rate``, which is *not* stored -- it is resolved from
+        ``res.currency.rate`` on every read -- so correcting or back-filling a
+        rate after the fact silently moves them. Without this guard the next
+        write to touch the move's lines, reconciling a payment among them,
+        would re-book a posted entry and push the difference onto the payable.
+        """
+        for move in self:
+            if move.move_type not in ("in_invoice", "in_refund"):
+                continue
+            if move.state == "posted":
```
yostashiro:
```suggestion
            if move.is_purchase_document() or move.state == "posted":
```

yostashiro (commented):
Partial review and it's taking a lot of time.

Can we shorten/remove comments and docstrings as they seem to be overly verbose and sometimes incorrect.

I think it's a good candidate for OCA. Is there something holding you back?

AungKoKoLin1997:
Followed up on suggestion points and created a OCA PR.
https://github.com/OCA/purchase-workflow/pull/3161

